PR #125, opened 2015-09-15
Use Data.Void from base

int-index:
fixes #99

PR #129, opened 2015-10-12
Bump lower bound for base, to avoid trying to build 2.x with 7.8.x.

romanb:
At least since https://github.com/goldfirere/singletons/pull/125, singletons 2.x no longer compiles with 7.8 but the current lower bound for `base` still results in picking a build plan with 7.8 and singletons 2.x which is doomed to fail. Hence I hereby suggest to bump the lower bound for `base`, in order to avoid build plans being selected with 7.8 and singletons 2.x.

### DIFF
--- a/singletons.cabal
+++ b/singletons.cabal
@@ -42,7 +42,7 @@ source-repository this
 
 library
   hs-source-dirs:     src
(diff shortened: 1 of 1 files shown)


goldfirere:
Of course. Thanks!

I will also update the packages on Hackage.

PR #133, opened 2015-11-15
Depend on base >=4.8.0.0 && <5

k0001:
It builds fine with GHC 7.10.2 and GHCJS. 

GHCJS is still using `base-4.8.0.0`, that is why I suggest this change.

### DIFF
--- a/singletons.cabal
+++ b/singletons.cabal
@@ -42,7 +42,7 @@ source-repository this
 
 library
   hs-source-dirs:     src
-  build-depends:      base >= 4.8.1.0 && < 5,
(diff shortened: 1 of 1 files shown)


goldfirere:
But I believe `base-4.8.0.0` corresponds with GHC 7.10.1, for which the testsuite fails. That's the reason for the version restriction. Do you find that the testsuite passes with GHCJS?

(Aside: I'm surprised that Travis finds a failure with this patch!)

goldfirere:
This looks abandoned by the original author. Please re-post if you are still interested in this change. Thanks!

PR #154, opened 2016-07-01
Explicit forall in withSomeSing

### LINKED ISSUE #153: Question about definition of Sing
Hi, `Sing` is described as a "kind-idexed" data family, and I'm wondering why it's defined as

```haskell
data family Sing (a :: k) :: Type

data instance Sing (a :: Maybe k) where
  SNothing :: Sing Nothing
  SJust    :: Sing k -> Sing (Just k)
```

rather than

```haskell
data family Sing k :: k -> Type

data instance Sing (Maybe k) x where
  SNothing :: Sing (Maybe k) Nothing
  SJust    :: Sing k a -> Sing (Maybe k) (Just a)
```

Are these effectively the same type?

int-index:
closes #153

### DIFF
--- a/src/Data/Singletons.hs
+++ b/src/Data/Singletons.hs
@@ -283,7 +283,8 @@ withSingI sn r =
 
 -- | Convert a normal datatype (like 'Bool') to a singleton for that datatype,
 -- passing it into a continuation.
-withSomeSing :: SingKind k
+withSomeSing :: forall k r
+              . SingKind k
              => DemoteRep k                       -- ^ The original datatype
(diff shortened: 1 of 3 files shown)


goldfirere:
This is really a GHC bug. Your `forall` should be exactly what GHC infers. But it doesn't. I've reported this (to myself) as goldfirere/ghc#70, as I'm sure it's my bug to fix (and I didn't feel like making a minimal test case for a proper report to GHC central).

In any case, adding the forall certainly doesn't hurt.

PR #148, opened 2016-06-12
Remove KProxy

int-index:
Remove kind proxies from `singletons`.

PR #158, opened 2016-08-24
Remove lingering references to KProxy

RyanGlScott:
#148 removed uses of `KProxy` from the codebase, but it's still lurking in a couple of places:

* The `ByHand2` test
* An outdated comment about orphan instances in `Data.Promotion.Prelude.Enum`. Since it's outdated, I opted to remove the whole comment (and the unnecessary `-fno-warn-orphans` flag).
* The `README`. But see my inline comment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,18 +146,18 @@ generated for every new singleton type.
 A class used to pass singleton values implicitly. The `sing` method produces
 an explicit singleton value.
 
-    data SomeSing (kproxy :: KProxy k) where
-      SomeSing :: Sing (a :: k) -> SomeSing ('KProxy :: KProxy k)
+    data SomeSing k where
+      SomeSing :: Sing (a :: k) -> SomeSing k
 
 The `SomeSing` type wraps up an _existentially-quantified_ singleton. Note that
 the type parameter `a` does not appear in the `SomeSing` type. Thus, this type
 can be used when you have a singleton, but you don't know at compile time what
-it will be. `SomeSing ('KProxy :: KProxy Thing)` is isomorphic to `Thing`.
+it will be. `SomeSing Thing` is isomorphic to `Thing`.
 
-    class (kparam ~ 'KProxy) => SingKind (kparam :: KProxy k) where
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -146,18 +146,18 @@ generated for every new singleton type.
...
-      type DemoteRep kparam :: *
-      fromSing :: Sing (a :: k) -> DemoteRep kparam
-      toSing   :: DemoteRep kparam -> SomeSing kparam
+    class SingKind k where
+      type DemoteRep k :: *
+      fromSing :: Sing (a :: k) -> DemoteRep k
+      toSing   :: DemoteRep k -> SomeSing k
 
 This class is used to convert a singleton value back to a value in the
 original, unrefined ADT. The `fromSing` method converts, say, a
```
RyanGlScott: Further down, this says:

>  The `DemoteRep` associated
> kind-indexed type family maps a proxy of the kind `Nat`
> back to the type `Nat`.

This is no longer true, but I'm not sure what I should change this to. Should I just say "maps `Nat` to `Nat`"? I don't know if that really elucidates what the purpose of `DemoteRep` is. Maybe we ought to give an example where `DemoteRep k /= k`? (Speaking of which, should the default definition of `DemoteRep` be `DemoteRep k = k`?)

RyanGlScott:
Urk, the Travis build failed due to `tasty-golden` failing to build with `optparse-applicative-0.13`: https://github.com/feuerbach/tasty/issues/145#issuecomment-242058799

README.md
int-index: The obvious example where `DemoteRep k /= k` is `TyFun` with `->`.

RyanGlScott: True. However, I just realized that it still makes sense to distinguish between the kind-level `Nat` and the type-level `Nat` (between which `DemoteRep` facilitates the transition), so I simply updated the description to state that.

RyanGlScott:
There we go, the build is green now. All good to go.

goldfirere:
Hooray!

PR #169, opened 2016-11-30
Remove Proxy in KindOf

int-index:
`KindOf` was defined in terms of `Proxy` because it was (mis)used in `KindInference` constructors. Now there's a dedicated type for kind unification, `SameKind`.

### DIFF
--- a/src/Data/Singletons.hs
+++ b/src/Data/Singletons.hs
@@ -1,7 +1,7 @@
 {-# LANGUAGE MagicHash, RankNTypes, PolyKinds, GADTs, DataKinds,
              FlexibleContexts, FlexibleInstances,
              TypeFamilies, TypeOperators, TypeFamilyDependencies,
-             UndecidableInstances, TypeInType #-}
+             UndecidableInstances, TypeInType, ConstraintKinds #-}
 
 -----------------------------------------------------------------------------
 -- |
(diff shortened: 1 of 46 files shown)


src/Data/Singletons.hs
```diff
@@ -68,7 +68,7 @@ import GHC.Exts ( Proxy# )
 
 -- | Convenient synonym to refer to the kind of a type variable:
 -- @type KindOf (a :: k) = ('Proxy :: Proxy k)@
```
goldfirere: Please update the comment, too!

src/Data/Singletons/Util.hs
```diff
@@ -29,6 +30,13 @@ import Data.Foldable
 import Data.Traversable
 import Data.Generics
 import Control.Monad.Fail ( MonadFail )
+import GHC.Exts (Constraint)
+
+-- | Force GHC to unify the kinds of @a@ and @b@. Note that @SameKind a b@ is
+-- different from @KindOf a ~ KindOf b@ in that the former makes the kinds
+-- unify immediately, whereas the latter is a proposition that GHC considers
+-- as possibly false.
+type SameKind (a :: k) (b :: k) = (() :: Constraint)
```
goldfirere: I think this belongs better in `Data.Singletons`, as it will appear in the generated code.

int-index:
Thanks for the review, I updated the PR according to your comments.

int-index:
@goldfirere Do you want to see any other changes before merging?

goldfirere:
Hooray! Thanks!

Issue #149, opened 2016-06-13
Remove Proxy from promoted classes


goldfirere:
In GHC8, there should be no reason to use `Proxy` in a promoted class. But efforts at removing this proxy have been stymied. I've not investigated fully, but here is what I know:
- The problem seems unrelated to TH. When I copied spliced code back into the buffer, the error did not change.
- There's a chance that the error is due to CUSKiness: there's a definition which looks CUSKy but I want non-CUSK inference.
- Further probing led to goldfirere/ghc#69. So I gave up. But I doubt that's the real reason for the error.

In any case, I'm pretty sure this is blocked upstream.

Issue #145, opened 2016-05-03
Is it possible to promote higher-kinded type classes?


ocharles:
The following

``` haskell
$(promote [d|
  class Col f where
    col :: f a -> a -> Bool
  |])
```

results in

```
       Cannot promote an applied type variable f_1627517956.
```

Is this going to be impossible to express with `singletons`?

goldfirere:
Yes, this should be possible, but it's not right now -- this is an improvement that GHC 8 permits but `singletons` doesn't yet take advantage of.

To make this work, you'd need to remove `KProxy` in its entirety from `singletons`. This should be fine in GHC 8, because the poly-kinded `Proxy` can be promoted. Once everything is in terms of `Proxy` instead of `KProxy`, then you should be able to simply remove the check for promoting an applied type variable and it should work.

Happy to take a PR. :)

ocharles:
Great! This sounds like a fun bit of work, I'll see what I can come up with.

int-index:
With `TypeInType`, why use `Proxy` at all?

goldfirere:
It's not `TypeInType` that obviates `Proxy`, it's `TypeApplications`. The problem is with eliminating `Proxy` is still twofold:
1. Template Haskell does not support visible type applications.
2. GHC still lacks support for visible type patterns, meaning that proxies are still sometimes necessary in data constructors. There's no theoretical challenge with visible type patterns, but the design isn't obvious; nor is the engineering. See [GHC#11350](https://ghc.haskell.org/trac/ghc/ticket/11350).

So, for now, we still need `Proxy`. But we shouldn't need `KProxy`!

int-index:
`TypeApplications` eliminate a different use case for `Proxy`.

I meant that instead of replacing `SingKind (kproxy :: KProxy k)` with `SingKind (proxy :: Proxy k)` we can replace it with `SingKind k`, eliminating proxies altogether.

PR #171, opened 2016-12-06
Promoted classes without Proxy

int-index:
closes https://github.com/goldfirere/singletons/issues/149

In the light of https://ghc.haskell.org/trac/ghc/ticket/12928, I decided to proceed as follows:

* if a class type variable has no explicit kind, we make no effort to guess it and default to `*`. This is OK because before `TypeInType` we were limited by `KProxy` anyway.
* if a class type variable has an explicit kind, it is preserved.

This way, we always get CUSKs where we need them.

I also modified `promoteType` slightly, so this PR closes https://github.com/goldfirere/singletons/issues/145, because now this code compiles:

```
{-# LANGUAGE TypeInType, ScopedTypeVariables, TypeFamilies, GADTs, TemplateHaskell #-}

import Data.Kind
import Data.Singletons.Promote

$(promote [d|
  class Column (f :: Type -> Type) where
    col :: f a -> a -> Bool
  |])
```

### DIFF
--- a/src/Data/Singletons/Promote/Type.hs
+++ b/src/Data/Singletons/Promote/Type.hs
@@ -29,10 +29,8 @@ promoteType = go []
     go args     (DAppT t1 t2) = do
       k2 <- go [] t2
       go (k2 : args) t1
-    go args     (DSigT ty _) = go args ty  -- just ignore signatures
-    go []       (DVarT name) = return $ DVarT name
-    go _        (DVarT name) = fail $ "Cannot promote an applied type variable " ++
-                                      show name ++ "."
+    go args     (DSigT ty ki) = DSigT <$> go args ty <*> go [] ki
+    go args     (DVarT name) = return $ foldType (DVarT name) args
     go []       (DConT name)
       | name == typeRepName               = return DStarT
       | name == stringName                = return $ DConT symbolName
--- a/src/Data/Singletons/Promote/Type.hs
+++ b/src/Data/Singletons/Promote/Type.hs
@@ -29,7 +29,10 @@ promoteType = go []
     go args     (DAppT t1 t2) = do
       k2 <- go [] t2
       go (k2 : args) t1
-    go args     (DSigT ty ki) = DSigT <$> go args ty <*> go [] ki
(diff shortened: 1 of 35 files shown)


int-index:
Just to make it clear, this version does *not* remove `KindInference` constructors.

src/Data/Singletons/Promote/Type.hs
```diff
@@ -29,10 +29,8 @@ promoteType = go []
     go args     (DAppT t1 t2) = do
       k2 <- go [] t2
       go (k2 : args) t1
-    go args     (DSigT ty _) = go args ty  -- just ignore signatures
-    go []       (DVarT name) = return $ DVarT name
-    go _        (DVarT name) = fail $ "Cannot promote an applied type variable " ++
-                                      show name ++ "."
+    go args     (DSigT ty ki) = DSigT <$> go args ty <*> go [] ki
```
goldfirere: This seems wrong. It will treat `(x :: k) y` as `(x y :: k)`, I think.

src/Data/Singletons/Single/Monad.hs
```diff
@@ -228,3 +228,6 @@ singDecsM :: DsMonad q => [Dec] -> SgM [DDec] -> q [DDec]
 singDecsM locals thing = do
   (decs1, decs2) <- singM locals thing
   return $ decs1 ++ decs2
+
+proxyFor :: DType -> DExp
```
goldfirere: The one use of this function will be able to removed once goldfirere/th-desugar#47 is merged, I think. I don't love the one-off use of the `Proxy` names here, but it should be temporary enough.

int-index: You are right, I intended this function to be temporary, which is why it's in an inappropriate module and uses `Proxy` names this way. Looking forward to getting rid of it.

src/Data/Singletons/Promote/Type.hs
int-index: To be honest, I didn't entirely understand what I was doing here. I'll add a test.

goldfirere:
Yay! Thanks!

PR #181, opened 2017-03-04
Combine clauses of record-selector function declarations

### LINKED ISSUE #180: Error with promotion fields for sum types
When I write
```
promote [d|
  data X = X1 {y :: Symbol} | X2 {y :: Symbol}
  z (X1 x) = x
  z (X2 x) = x
  |]
```
I expect that Y and Z are similar. But now definition for Y is wrong:

```
:i Y
type family Y (a0 :: X) :: Symbol
  where [(field0 :: Symbol)] Y ('X2 field0) = field0
:i Z
type family Z (a0 :: X) :: Symbol
  where
    [(x0 :: Symbol)] Z ('X1 x0) = x0
    [(x0 :: Symbol)] Z ('X2 x0) = x0
```
 That is on [stack lts-8.0](https://www.stackage.org/lts-8.0/package/singletons-2.2)

RyanGlScott:
Previously, in `promoteDataDecs` we were simply calling `getRecordSelectors` on each constructor of a data type and building a `LetDecEnv` map of names to function declarations clauses. But if a record selector is contained in multiple constructors, then the clauses of the record selector function will be spread out across multiple `DLetD`s, so when the `LetDecEnv` is built, it only keeps the most recent `DLetD` for a particular record selector name! This causes the incorrect promoted definition that appeared in #180.

This adds `mergeLetDecs` to post-process the results of `getRecordSelectors` so that all of the clauses for a particular record selector are grouped together before being turned into a `LetDecEnv`. I've added a lengthy comment above `mergeLetDecs` explaining what it does and why it's needed.

Fixes #180.

### DIFF
--- a/src/Data/Singletons/Promote.hs
+++ b/src/Data/Singletons/Promote.hs
@@ -27,6 +27,7 @@ import Data.Singletons.Util
 import Data.Singletons.Syntax
 import Prelude hiding (exp)
 import Control.Monad
+import Data.List (partition)
 import qualified Data.Map.Strict as Map
 import Data.Map.Strict ( Map )
 import Data.Maybe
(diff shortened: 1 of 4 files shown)


src/Data/Singletons/Promote.hs
```diff
@@ -189,7 +190,55 @@ promoteDataDecs data_decs = do
       let arg_ty = foldType (DConT data_name)
                             (map tvbToType tvbs)
       in
-      concatMapM (getRecordSelectors arg_ty) cons
+      mergeLetDecs <$> concatMapM (getRecordSelectors arg_ty) cons
+
+-- After retrieving the record selectors from a data type's constructors, it
```
goldfirere: Great comment.

src/Data/Singletons/Promote.hs
```diff
@@ -189,7 +190,55 @@ promoteDataDecs data_decs = do
...
+-- which the record selector is present.
+mergeLetDecs :: [DLetDec] -> [DLetDec]
+mergeLetDecs [] = []
+mergeLetDecs (x:xs)
+  -- If we encounter a function declaration, looks for all other function
+  -- declarations in the rest of the list with the same name, and concat
+  -- their clauses.
+  | DFunD n clauses <- x
+  = let (eq_n, neq_n)  = partition (\case DFunD n2 _ -> n == n2; _ -> False) xs
+        merged_clauses = concat $ clauses:map (\(DFunD _ cls) -> cls) eq_n
```
goldfirere: `partitionWith` (stolen from GHC and put in `Data.Singletons.Util`) should simplify this.

RyanGlScott:
I agree that making `getRecordSelectors` work on one `Con` at a time is fraught with peril. I'd be fine with removing `getRecordSelectors` from `th-desugar` as well—where in `singletons` would be the best home for it?

src/Data/Singletons/Promote.hs
RyanGlScott: Good catch, I'll refactor it to use `partitionWith`.

goldfirere:
Considering `getRecordSelectors` is used only in Promote, that looks like the best spot for the new functionality. Thanks!

RyanGlScott:
Oh, it turns out there'd be one awkward aspect of removing `getRecordSelectors` from `th-desugar` entirely—it's used in [the test suite](https://github.com/goldfirere/th-desugar/blob/92ff81db52758a3131f4fdb1398204ad1be617ac/Test/DsDec.hs#L74). I'm not sure what the best way to deal with this is... should we:

* Find a way to alter that test?
* Keep `getRecordSelectors` in `th-desugar`, but port it over to `singletons` with the necessary fixes?
* Keep `getRecordSelectors` in `th-desugar`, and put the necessary fixes in place there?

goldfirere:
I suppose option 3 is the most widely helpful.... just in case some user of `th-desugar` (other than `singletons`) uses `getRecordSelectors`.

RyanGlScott:
Closing in favor of https://github.com/goldfirere/th-desugar/issues/55.

PR #182, opened 2017-04-22
WIP: GHC 8.2 revamp

### LINKED ISSUE #180: Error with promotion fields for sum types

RyanGlScott:
Fixes #180 while I'm in town.


src/Data/Singletons/Promote.hs
```diff
@@ -552,6 +554,7 @@ promotePat (DBangPa pat) = do
   qReportWarning "Strict pattern converted into regular pattern in promotion"
   (ty, pat') <- promotePat pat
   return (ty, DBangPa pat')
+promotePat (DSigPa _pat _ty) = error "TODO: Handle SigPa"
```
RyanGlScott: To be honest, I'm not quite sure how one would promote a `DSigPa`. Suggestions welcome.

src/Data/Singletons/Single.hs
```diff
@@ -499,6 +501,7 @@ singPat var_proms patCxt (DTildePa pat) = do
 singPat var_proms patCxt (DBangPa pat) = do
   (pat', ty) <- singPat var_proms patCxt pat
   return (DBangPa pat', ty)
+singPat _var_proms _patCxt (DSigPa _pat _ty) = error "TODO: Handle SigPa"
```
RyanGlScott: Similarly, how do we single `DSigPa`s?

src/Data/Singletons/Partition.hs
```diff
@@ -46,23 +48,35 @@ partitionDecs :: Quasi m => [DDec] -> m PartitionedDecs
...
+         | stock, deriv_name == boundedName
         -> Right <$> mkBoundedInstance ty cons
-         | deriv_name == enumName
+         | stock, deriv_name == enumName
         -> Right <$> mkEnumInstance ty cons
+        where
+          stock = case strat of
+                    Nothing            -> True
+                    Just StockStrategy -> True
+                    Just _             -> False
```
RyanGlScott: I opted to only choose `deriving` clauses with an explicit strategy when that strategy is `stock`. We could make it more permissive and ignore the strategy, however.

src/Data/Singletons/Prelude/List.hs
```diff
@@ -306,12 +306,18 @@ $(singletonsOnly [d|
...
     where
       perms []     _  = []
       perms (t:ts) is = foldr interleave (perms ts (t:is)) (permutations is)
         where interleave    xs     r = let (_,zs) = interleave' id xs r in zs
+
+              -- This type signature isn't present in the reference
+              -- implementation of permutations in base. However, it is needed
+              -- here, since (at least in GHC 8.2.1) the singletonized version
+              -- will fail to typecheck without it. See #13549 for the full story.
+              interleave' :: ([a] -> b) -> [a] -> [b] -> ([a], [b])
```
RyanGlScott: It turns out adding one type signature is enough to work around [GHC Trac #13549](https://ghc.haskell.org/trac/ghc/ticket/13549). Hooray!

tests/compile-and-dump/Singletons/BadBoundedDeriving.ghc82.template
```diff
@@ -1,3 +1,6 @@
 
 Singletons/BadBoundedDeriving.hs:0:0: error:
     Can't derive Bounded instance for Foo_0 a_1.
+  |
+6 | $(singletons [d|
+  |   ^^^^^^^^^^^^^^...
```
RyanGlScott: Do you think we should include these error message locations in golden files? They do add some degree of bloat.

RyanGlScott:
Another TODO: the Travis build is failing because `Cabal-2.0` is inclined to build the test suite _before_ the library. Unfortunately, this breaks the assumption that `"../dist/build/autogen/cabal_macros.h"` will exist by the time the test suite is built (see [this Travis error](https://travis-ci.org/goldfirere/singletons/jobs/224782539#L656) for an example).

Perhaps we should just make the test suite depend on `singletons` in `singletons.cabal`? Then again, there's [this Note](https://github.com/goldfirere/singletons/blob/f6516af95f2421f9509abfb992523d430e00b6ad/tests/SingletonsTestSuiteUtils.hs#L121-L124):

```
-- Note [-this-unit-id hack]
-- ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--
-- We want to avoid installing singletons package before running the
-- testsuite, because in this way we prevent double compilation of the
-- library.
```

I don't think making the test suite depend on `singletons` in `singletons.cabal` would result in double compilation, but then again, I am quite new to the inner machinations of this test suite.

src/Data/Singletons/Prelude/List.hs
goldfirere: See my comment to GHC's #13459. This type signature is indeed necessary: a type signature is required on a local function if its type is inferred differently depending on the presence of `MonoLocalBinds`. This is the case here.

goldfirere:
Somehow just saw your last comment, about Cabal and `-this-unit-id`. It's quite possible that modern Cabals won't cause double-compilation.... or simply that we had configured it wrongly when setting this up. I don't think it would severely break things to add the `singletons` dependency.

RyanGlScott:
OK, adding `singletons` as a test suite dependency seemed to fix that particular problem. But now the tests themselves are failing, apparently because of different unique values. But we enable `-dsuppress-uniques`... hm. I need to stare at this more closely.

RyanGlScott:
Ah, here's a hunch. In the test suite, we do [normalization with `sed`](https://github.com/goldfirere/singletons/blob/023513fe2ee63391095540f575873064bbe3ab03/tests/SingletonsTestSuiteUtils.hs#L199-L218):

```
-- Note [Normalization with sed]
-- ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--
-- Output file is normalized with sed. Line numbers generated in splices:
--
--   Foo:(40,3)-(42,4)
--   Foo.hs:7:3:
--   Equals_1235967303
--
-- are turned into:
--
--   Foo:(0,0)-(0,0)
--   Foo.hs:0:0:
--   Equals_0123456789
```

In particular, we normalize [this pattern](https://github.com/goldfirere/singletons/blob/023513fe2ee63391095540f575873064bbe3ab03/tests/SingletonsTestSuiteUtils.hs#L230):

```haskell
"-e", "'s/[0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9]/0123456789/g'"
```

This works assuming that whenever you have something like `Equals_1235967303`, it will always have exactly 10 numbers following it. But in GHC 8.2, uniques appear to have 19 numbers (perhaps due to [this commit](http://git.haskell.org/ghc.git/commit/0d213c18b6962bb65e2b3035a258dd3f5bf454dd)?). Maybe adding a pattern for 19 numbers will do the trick.

RyanGlScott:
Luckily, adding more patterns for `sed` was precisely the secret sauce needed to cook up a green Travis build.

RyanGlScott:
Should this be closed now that 637fb4d0435220f740bc658973f7a05a499dcee6 has been committed?

goldfirere:
Don't close quite yet. I did merge (it passes the testsuite) but I haven't reviewed carefully yet. Will do so soon.

goldfirere:
This has been merged.

PR #185, opened 2017-06-01
Banish some more Proxies

RyanGlScott:
* The `ByHand` tests no longer compile after commit 444db17856018555ed1e9c89cc16f6d6ed0795a9, so let's update them to avoid `Proxy` when appropriate.
* The `SuppressUnusedWarnings` class doesn't really _need_ to use `Proxy`, does it? Let's remove it from there as well.

### DIFF
--- a/tests/ByHand2.hs
+++ b/tests/ByHand2.hs
@@ -7,7 +7,6 @@ module ByHand2 where
 
 import Prelude hiding ( Eq(..), Ord(..) )
 import Data.Singletons
-import Data.Proxy
 
 data Nat = Zero | Succ Nat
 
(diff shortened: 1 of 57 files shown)


tests/ByHand2.hs
```diff
@@ -201,5 +200,6 @@ instance SFD Bool Nat where
   sL2r STrue = SSucc SZero
 
 sT1 = sMeth STrue
-sT2 :: Sing T2
+-- We can't use this type signature due to GHC Trac #13774
```
RyanGlScott: ARGH

goldfirere:
Good catch. Thanks.

PR #186, opened 2017-06-01
Fix haddock failures for TH splices

christiaanb:
On my machine, using ghc-8.2.0.20170522, I was getting many:
"Haddock documentation not (yet) handled by Template Haskell"

Maybe it's my machine/config?


goldfirere:
I haven't tried running haddock... so this problem is probably more than just your machine. Thanks.

PR #188, opened 2017-06-02
Fix #187

### LINKED ISSUE #187: "Impossible! A function without clauses" when deriving Ord for empty datatype
I decided to be cruel and asked `singletons` to derive `Ord` for an empty datatype:

```haskell
import Data.Singletons.TH

$(singletons [d| data Foo deriving (Eq, Ord) |])
```
```
GHCi, version 8.2.0.20170522: http://www.haskell.org/ghc/  :? for help
[1 of 1] Compiling Main             ( Bug.hs, interpreted )

Bug.hs:4:3: error: Impossible! A function without clauses.
  |
4 | $(singletons [d| data Foo deriving (Eq, Ord) |])
  |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Bug.hs:4:3: error: Q monad failure
  |
4 | $(singletons [d| data Foo deriving (Eq, Ord) |])
  |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Whereas the error messages you get when deriving `Bounded` and `Enum` for empty datatypes aren't as alarming:

```
λ> ; $(singletons [d| data Foo deriving Enum |])

<interactive>:7:5: error: Can't derive Enum instance for Foo_0.
λ> ; $(singletons [d| data Foo deriving Bounded |])

<interactive>:8:5: error: Can't derive Bounded instance for Foo_0.
```

RyanGlScott:
Allowing derived `Ord` instances for empty datatypes proves to be not so challenging, as we can effectively generate this implementation of `compare` (promoted and singled, of course):

```haskell
compare _ _ = foldl thenCmp EQ []
```

Fixes #187.

### DIFF
--- a/src/Data/Singletons/Deriving/Ord.hs
+++ b/src/Data/Singletons/Deriving/Ord.hs
@@ -24,36 +24,27 @@ import Data.Singletons.Syntax
 mkOrdInstance :: Quasi q => DType -> [DCon] -> q UInstDecl
 mkOrdInstance ty cons = do
   let constraints = inferConstraints (DConPr ordName) cons
-  -- This case also handles empty datatypes, since we can just create
-  -- one clause with wildcard matches and zero fields to fold over
-  compare_eq_clauses <- mapM mk_equal_clause $ if null cons
-                                                  then [Nothing]
-                                                  else map Just cons
+  compare_eq_clauses <- mapM mk_equal_clause cons
   let compare_noneq_clauses = map (uncurry mk_nonequal_clause)
                                   [ (con1, con2)
                                   | con1 <- zip cons [1..]
                                   , con2 <- zip cons [1..]
                                   , extractName (fst con1) /=
                                     extractName (fst con2) ]
+      clauses | null cons = [mk_empty_clause]
+              | otherwise = compare_eq_clauses ++ compare_noneq_clauses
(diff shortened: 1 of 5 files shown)


src/Data/Singletons/Deriving/Ord.hs
```diff
@@ -24,7 +24,11 @@ import Data.Singletons.Syntax
 mkOrdInstance :: Quasi q => DType -> [DCon] -> q UInstDecl
 mkOrdInstance ty cons = do
   let constraints = inferConstraints (DConPr ordName) cons
-  compare_eq_clauses <- mapM mk_equal_clause cons
+  -- This case also handles empty datatypes, since we can just create
+  -- one clause with wildcard matches and zero fields to fold over
+  compare_eq_clauses <- mapM mk_equal_clause $ if null cons
```
goldfirere: The use of `Maybe` here seems a bit off. Instead, I would keep this (and the next line, for `compare_noneq_clauses`) as they are, and instead just generate a different `UFunction` at the last step. Seems simpler.

RyanGlScott:
How does it look now?

goldfirere:
Much better. Thanks.

PR #189, opened 2017-06-02
Promote wildcard patterns to wildcard types

RyanGlScott:
GHC 8.0 and later support wildcard types, so we no longer have to generate fresh names to promote wildcard patterns. This has the nice benefit of making the generated code slightly less noisy.


RyanGlScott:
There's actually another source of unnecessary fresh names for wildcard patterns: the [`removeWilds`](https://github.com/goldfirere/th-desugar/blob/3ee50366f99b6183b6b4da3cca1d272a8edfb226/Language/Haskell/TH/Desugar/Core.hs#L690-L699) function in `th-desugar`, which is used when [desugaring as-patterns](https://github.com/goldfirere/th-desugar/blob/3ee50366f99b6183b6b4da3cca1d272a8edfb226/Language/Haskell/TH/Desugar/Core.hs#L655). However, I don't quite understand _why_ `removeWilds` is used there...

goldfirere:
Yes, looks good. Thanks!

As to `removeWilds`, that looks necessary. th-desguar transforms `f x@(_:xs) = blah` to become `f (_fresh:xs) = let x = _fresh:xs in blah`, so we need a proper `_fresh` name.

Issue #190, opened 2017-06-05
"Cannot find" errors when deriving due to a lack of imports


RyanGlScott:
There's a bit of discrepancy among the derivable type classes when it comes to what you need to import to actually make them work. To explain what I mean, consider this example:

```haskell
{-# LANGUAGE ExistentialQuantification #-}
{-# LANGUAGE GADTs #-}
{-# LANGUAGE InstanceSigs #-}
{-# LANGUAGE ScopedTypeVariables #-}
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE TypeInType #-}
{-# LANGUAGE UndecidableInstances #-}
module Bug where

import Data.Singletons.TH

$(singletons [d| data T = T deriving Eq |])
```

This compiles, needing only an import of `Data.Singletons.TH`. If you try `Enum` instead, however:

```haskell
$(singletons [d| data T = T deriving Enum |])
```

You're greeted with this rather confusing error message:

```
Bug.hs:12:3: error:
    Cannot find class declaration annotation for Data.Singletons.Prelude.Enum.Enum
   |
12 | $(singletons [d| data T = T deriving Enum |])
   |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Bug.hs:12:3: error: Q monad failure
   |
12 | $(singletons [d| data T = T deriving Enum |])
   |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Similarly with `Bounded`:

```haskell
$(singletons [d| data T = T deriving Bounded |])
```
```
Bug.hs:12:3: error: Cannot find type annotation for MinBound
   |
12 | $(singletons [d| data T = T deriving Bounded |])
   |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Bug.hs:12:3: error: Q monad failure
   |
12 | $(singletons [d| data T = T deriving Bounded |])
   |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

`Ord` also gives an error (although not quite as confusing):

```haskell
$(singletons [d| data T = T deriving (Eq, Ord) |])
```
```
Bug.hs:12:3: error:
    • Data constructor not in scope: SEQ :: Sing 'EQ
[...]

Issue #160, opened 2016-09-24
Promote `error` to custom type error


crockeea:
With the addition of custom type errors, it should be possible to promote calls to `error` to a custom type error. For example, (very roughly)

```
singletons [d|
  foo :: Int -> Int
  foo 0 = 1
  foo _ = error $ "Can't call foo on " ++ (show x)
  |]
```

should get a type family

```
type family FFoo x where
  FFoo Z = S Z
  FFoo x = TypeError (Text "Can't call foo on " :<>: ShowType x)
```

It'd be great to change the `foo` in the error message to `FFoo`; any ideas how that can happen?

goldfirere:
I had initially thought this was easy, but now I'm backing off.

It would be easy to take `error "blah"` and promote that to `TypeError (Text "blah")`, but now that you introduce function calls, etc., it's getting harder. This effectively requires that `String`s be promoted either to `Symbol`s or to `ErrorMessage`s, depending on the context. This would be hard.

I think the Right Way to do this is to promote `Char` to `Symbol` (where each such `Symbol` is one-character long) and then `String` to `[Symbol]`. Then we can concatenate type-level strings. `error` can be treated as `TypeError` composed with `Text`. Then it all should work. But this is not a quick fix, I'm afraid.

RyanGlScott:
One other avenue we could explore is using `base-4.10.0.0`'s new [`AppendSymbol` type family](http://git.haskell.org/ghc.git/blob/69d9081d9fa3ff36fda36e4e11ef7e8f946ecf2a:/libraries/base/GHC/TypeLits.hs#l140) to append `Symbols` together.

goldfirere:
But how do we know whether to promote `++` to `:++` or to `AppendSymbol`?

Really, the problem is that `Symbol` is *not* a promoted string. It is Something Else.

RyanGlScott:
Perhaps we shouldn't be using `(++)` at all for appending `Symbol`s. They are abstract types, after all. I'm envisioning adding `Monoid` to `singletons` ([as you do in the `monads` branch](https://github.com/goldfirere/singletons/commit/d09004ef25590cb0748f67842c1549d1f3c10cab#diff-5ba76c7731f041b351bdcd590756fd1eR120)) and promoting/singling a `Monoid Symbol` instance that uses `AppendSymbol` as `mappend`. Then, you can promote this code:

```haskell
singletons [d|
  foo :: Int -> Int
  foo 0 = 1
[...]

PR #191, opened 2017-06-05
Add promotion, singling, and deriving for Show

RyanGlScott:
Now that `GHC.TypeLits` features an `AppendSymbol` type family, it's possible to promote/single 90% of the `Show` type class and its features. (For the 10% that isn't covered, see my comment above the `Show Symbol` instance, as well as the comment explaining the lack of a `PShow Nat` instance.)

This also adds support for `deriving Show`, along with some test cases to stress-test it.

### DIFF
--- a/singletons.cabal
+++ b/singletons.cabal
@@ -44,9 +44,10 @@ library
   hs-source-dirs:     src
   build-depends:      base >= 4.10 && < 5,
                       mtl >= 2.1.2,
+                      ghc-boot-th,
                       template-haskell,
                       containers >= 0.5,
-                      th-desugar >= 1.7 && < 1.8,
+                      th-desugar >= 1.7.1 && < 1.8,
                       syb >= 0.4,
                       text >= 1.2
   default-language:   Haskell2010
@@ -70,6 +71,7 @@ library
                       Data.Singletons.Prelude.List.NonEmpty,
                       Data.Singletons.Prelude.Maybe,
                       Data.Singletons.Prelude.Num
+                      Data.Singletons.Prelude.Show,
                       Data.Singletons.Prelude.Tuple,
                       Data.Promotion.Prelude,
                       Data.Promotion.TH,
@@ -84,6 +86,7 @@ library
                       Data.Promotion.Prelude.List.NonEmpty,
                       Data.Promotion.Prelude.Maybe,
                       Data.Promotion.Prelude.Num,
+                      Data.Promotion.Prelude.Show,
                       Data.Promotion.Prelude.Tuple,
                       Data.Singletons.TypeLits,
                       Data.Singletons.Decide,
@@ -93,6 +96,7 @@ library
                       Data.Singletons.Deriving.Bounded,
                       Data.Singletons.Deriving.Enum,
                       Data.Singletons.Deriving.Ord,
+                      Data.Singletons.Deriving.Show,
                       Data.Singletons.Prelude.List.NonEmpty.Internal,
                       Data.Singletons.Promote,
                       Data.Singletons.Promote.Monad,
--- a/singletons.cabal
+++ b/singletons.cabal
@@ -47,7 +47,7 @@ library
                       ghc-boot-th,
                       template-haskell,
                       containers >= 0.5,
-                      th-desugar >= 1.7.1 && < 1.8,
+                      th-desugar >= 1.8 && < 1.9,
                       syb >= 0.4,
                       text >= 1.2
   default-language:   Haskell2010
--- a/src/Data/Promotion/Prelude.hs
+++ b/src/Data/Promotion/Prelude.hs
@@ -42,6 +42,9 @@ module Data.Promotion.Prelude (
   -- * Promoted numbers
   module Data.Promotion.Prelude.Num,
 
+  -- * Promoted 'Show'
+  module Data.Promotion.Prelude.Show,
+
   -- ** Miscellaneous functions
   Id, Const, (:.), type ($), type ($!), Flip, AsTypeOf, Until, Seq,
 
@@ -161,4 +164,5 @@ import Data.Promotion.Prelude.Ord
 import Data.Promotion.Prelude.Enum
   hiding (Succ, Pred, SuccSym0, SuccSym1, PredSym0, PredSym1)
 import Data.Promotion.Prelude.Num
+import Data.Promotion.Prelude.Show
 import Data.Singletons.TypeLits
--- a/src/Data/Promotion/Prelude.hs
+++ b/src/Data/Promotion/Prelude.hs
@@ -43,7 +43,7 @@ module Data.Promotion.Prelude (
   module Data.Promotion.Prelude.Num,
 
   -- * Promoted 'Show'
-  PShow(..), ShowS, SChar, (:<>), Shows, ShowChar, ShowString, ShowParen,
+  PShow(..), ShowS, SChar, show_, (:<>), Shows, ShowChar, ShowString, ShowParen,
 
   -- ** Miscellaneous functions
   Id, Const, (:.), type ($), type ($!), Flip, AsTypeOf, Until, Seq,
@@ -98,7 +98,7 @@ module Data.Promotion.Prelude (
   (:^$), (:^$$),
 
   ShowsPrecSym0, ShowsPrecSym1, ShowsPrecSym2, ShowsPrecSym3,
-  Show'Sym0, Show'Sym1,
+  Show_Sym0, Show_Sym1,
   ShowListSym0, ShowListSym1, ShowListSym2,
   (:<>$), (:<>$$), (:<>$$$),
   ShowsSym0, ShowsSym1, ShowsSym2,
--- a/src/Data/Promotion/Prelude.hs
+++ b/src/Data/Promotion/Prelude.hs
@@ -43,7 +43,7 @@ module Data.Promotion.Prelude (
   module Data.Promotion.Prelude.Num,
 
   -- * Promoted 'Show'
-  module Data.Promotion.Prelude.Show,
+  PShow(..), ShowS, SChar, (:<>), Shows, ShowChar, ShowString, ShowParen,
 
   -- ** Miscellaneous functions
   Id, Const, (:.), type ($), type ($!), Flip, AsTypeOf, Until, Seq,
@@ -96,6 +96,15 @@ module Data.Promotion.Prelude (
   UncurrySym0, UncurrySym1, UncurrySym2,
 
   (:^$), (:^$$),
+
+  ShowsPrecSym0, ShowsPrecSym1, ShowsPrecSym2, ShowsPrecSym3,
+  Show'Sym0, Show'Sym1,
+  ShowListSym0, ShowListSym1, ShowListSym2,
+  (:<>$), (:<>$$), (:<>$$$),
+  ShowsSym0, ShowsSym1, ShowsSym2,
+  ShowCharSym0, ShowCharSym1, ShowCharSym2,
+  ShowStringSym0, ShowStringSym1, ShowStringSym2,
+  ShowParenSym0, ShowParenSym1, ShowParenSym2,
 
   IdSym0, IdSym1, ConstSym0, ConstSym1, ConstSym2,
   (:.$), (:.$$), (:.$$$),
--- a/src/Data/Promotion/Prelude/Show.hs
+++ b/src/Data/Promotion/Prelude/Show.hs
@@ -12,38 +12,23 @@
 -----------------------------------------------------------------------------
 
 module Data.Promotion.Prelude.Show (
-  PShow(..), SymbolS, SChar,
+  PShow(..), SymbolS, SChar, show_, (:<>),
+  Shows, ShowListWith, ShowChar, ShowString, ShowParen,
+  ShowSpace, ShowCommaSpace, AppPrec, AppPrec1,
 
+  -- * Defunctionalization symbols
   ShowsPrecSym0, ShowsPrecSym1, ShowsPrecSym2, ShowsPrecSym3,
-  Show'Sym0, Show'Sym1,
+  Show_Sym0, Show_Sym1,
   ShowListSym0, ShowListSym1, ShowListSym2,
-
-  (:<>),
   (:<>$), (:<>$$), (:<>$$$),
-
-  Shows,
   ShowsSym0, ShowsSym1, ShowsSym2,
-
-  ShowListWith,
   ShowListWithSym0, ShowListWithSym1, ShowListWithSym2, ShowListWithSym3,
-
-  ShowChar,
   ShowCharSym0, ShowCharSym1, ShowCharSym2,
-
-  ShowString,
   ShowStringSym0, ShowStringSym1, ShowStringSym2,
-
-  ShowParen,
   ShowParenSym0, ShowParenSym1, ShowParenSym2,
-
-  ShowSpace,
   ShowSpaceSym0, ShowSpaceSym1,
-
-  ShowCommaSpace,
   ShowCommaSpaceSym0, ShowCommaSpaceSym1,
-
-  AppPrec, AppPrecSym0,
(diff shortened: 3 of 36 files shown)


goldfirere:
Wow. This is quite a bit of work!

I hate to ask, but: Why? It has occurred to me to allow this kind of thing, but I always thought a singletonized version of `Show` was useless.

RyanGlScott:
My use case is type-level error messages, and works towards improving the story for #160. Currently, you have only primitive ways of constructing an `ErrorMessage`:

```haskell
data {-kind-} ErrorMessage = Text Symbol
                           | forall t. ShowType t
```

`ShowType` gives you one way of lifting arbitrary types to error messages, but it locks you into a particular way of rendering the types. There are a number of things about `ShowType`'s pretty-printing that I (and imagine others) would find unsavory, such as:

1. `ShowType` always renders promoted ticks, which can look messy:

   ```
   λ> let f :: TypeError (ShowType (Just True)); f = undefined
   
   <interactive>:11:10: error:
       • 'Just 'True
       • In the type signature:
           f :: TypeError (ShowType (Just True))
   ```
2. `ShowType` doesn't take fixity into account:

   ```
   λ> let f :: TypeError (ShowType (True `Infix` True)); f = undefined
   
   <interactive>:18:10: error:
       • Infix 'True 'True
       • In the type signature:
           f :: TypeError (ShowType (True `Infix` True))
   ```
3. More generally, `ShowType` is not _customizable_. The only way to make `ShowType` give the output you want to is carefully craft your types such that GHC's pretty-printer displays them the way you want, but that approach doesn't scale.

So, the other option is to use `Text`. But _prima facie_, there appears to be only one way to use `Text`: construct `Symbol`s by hand. But not with a promoted `PShow` class! `PShow` grants you the power to come up with flexible ways of turning your types into `Symbol`s in a composable way.

To really solve #160, we'll probably want to distinguish between `showType` (an `ErrorMessage` constructor) and `show'` (what this PR provides), as well as `(<>:)` (an `ErrorMessage` constructor) and `(<>)` (what this PR provides), for folks who want to use both, but I decided that was out of the scope of this PR.

goldfirere:
A compelling answer. I will review shortly.

src/Data/Singletons/Deriving/Show.hs
```diff
@@ -0,0 +1,140 @@
...
+                                   `DAppE` dStringE (parenInfixConName con_name " "))
+                         `DAppE` composed_args
+      return $ DClause [DVarPa p, DConPa con_name $ map DVarPa args] $
+        DVarE showParenName
+          `DAppE` (DVarE gtName `DAppE` DVarE p `DAppE` dIntegerE appPrec)
+          `DAppE` named_args
+
+    -- A record constructor with no fields? This probably won't ever happen,
+    -- but just to be safe, we'll act as if it were a DNormalC.
+    go (DRecC []) = go (DNormalC False [])
```
goldfirere: Yes, this can really happen. GHC's `Show` instance does not print the extra braces, just as yours behaves here.

src/Data/Singletons/Prelude.hs
```diff
@@ -54,6 +54,9 @@ module Data.Singletons.Prelude (
   -- * Singletons numbers
   module Data.Singletons.Prelude.Num,
 
+  -- * Singleton 'Show'
+  module Data.Singletons.Prelude.Show,
```
goldfirere: Ditto my comment above about the "promotion" Prelude.

src/Data/Singletons/Prelude/Show.hs
```diff
@@ -0,0 +1,229 @@
...
+{-# LANGUAGE TypeInType #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+{-# LANGUAGE TypeApplications #-}
+
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Data.Singletons.Prelude.Show
+-- Copyright   :  (C) 2013 Richard Eisenberg
```
goldfirere: You may as well put your own name here...

src/Data/Singletons/Prelude/Show.hs
```diff
@@ -0,0 +1,229 @@
...
+import qualified Data.Text as T
+
+import           GHC.TypeLits
+
+import qualified Prelude as P
+import           Prelude hiding (Show(..))
+
+import           Unsafe.Coerce (unsafeCoerce)
+
+-- | The promoted analogue of '(<>)' for 'Symbol's. This uses the special
```
goldfirere: I don't want to be forward-incompatible with `Data.Monoid`'s `(<>)`. But I suppose this isn't incompatible with it.

src/Data/Singletons/Prelude/Show.hs
```diff
@@ -0,0 +1,229 @@
...
+
+-- | GHC currently has no notion of type-level 'Char's, so we fake them with
+-- single-character 'Symbol's.
+type SChar = Symbol
+
+$(singletonsOnly [d|
+  class Show a where
+    showsPrec :: Nat -> a -> SymbolS
+    -- Intentionally renamed to avoid clashing with the Show class
+    show'     :: a -> Symbol
```
goldfirere: In other places, `singletons` uses a trailing underscore to avoid naming conflicts. See `Data.Singletons.Prelude.Maybe.maybe_`. But remember to export the term-level name, then, too.

src/Data/Singletons/Prelude/Show.hs
```diff
@@ -0,0 +1,229 @@
...
+  instance (Show a, Show b, Show c, Show d, Show e, Show f) => Show (a,b,c,d,e,f) where
+    showsPrec _ (a,b,c,d,e,f) s = show_tuple [shows a, shows b, shows c, shows d, shows e, shows f] s
+
+  instance (Show a, Show b, Show c, Show d, Show e, Show f, Show g)
+          => Show (a,b,c,d,e,f,g) where
+    showsPrec _ (a,b,c,d,e,f,g) s
+          = show_tuple [shows a, shows b, shows c, shows d, shows e, shows f, shows g] s
+  |])
+
+-- At the moment, I can't think of a way to implement a PShow Nat instance,
```
goldfirere: It can't be that hard... just write a normal `toString :: Integer -> String` and promote it, no? :)

src/Data/Singletons/Prelude/Show.hs
```diff
@@ -0,0 +1,229 @@
...
+
+  sShowList sl sx =
+    let l = fromSing sl
+        x = fromSing sx
+        ex = someSymbolVal (P.show l ++ T.unpack x)
+    in
+    case ex of
+      SomeSymbol (_ :: Proxy s) -> unsafeCoerce (SSym :: Sing s)
+
+$(singShowInstances [ ''(), ''Maybe, ''Either, ''NonEmpty, ''Bool, ''Ordering ])
```
goldfirere: Lists? Or is that handled by `showList`?

src/Data/Promotion/Prelude.hs
```diff
@@ -42,6 +42,9 @@ module Data.Promotion.Prelude (
   -- * Promoted numbers
   module Data.Promotion.Prelude.Num,
 
+  -- * Promoted 'Show'
+  module Data.Promotion.Prelude.Show,
```
RyanGlScott: Hm, good catch. I'll be more selective with what I export.

src/Data/Promotion/Prelude/Show.hs
```diff
@@ -0,0 +1,49 @@
...
+-- Maintainer  :  Jan Stolarek (jan.stolarek@p.lodz.pl)
+-- Stability   :  experimental
+-- Portability :  non-portable
+--
+-- Exports a promoted version of 'Show'
+--
+-----------------------------------------------------------------------------
+
+module Data.Promotion.Prelude.Show (
+  PShow(..), SymbolS, SChar,
```
RyanGlScott: Good catch. I'll do that.

singletons.cabal
```diff
@@ -44,9 +44,10 @@ library
   hs-source-dirs:     src
   build-depends:      base >= 4.10 && < 5,
                       mtl >= 2.1.2,
+                      ghc-boot-th,
                       template-haskell,
                       containers >= 0.5,
-                      th-desugar >= 1.7 && < 1.8,
+                      th-desugar >= 1.7.1 && < 1.8,
```
RyanGlScott: Good thinking. I'll bump it up.

src/Data/Singletons/Deriving/Show.hs
```diff
@@ -0,0 +1,140 @@
...
+import Data.Singletons.Syntax
+import Data.Singletons.Deriving.Infer
+import Control.Monad (when)
+import Data.Maybe (fromMaybe)
+import GHC.Lexeme (startsConSym, startsVarSym)
+import GHC.Show (appPrec, appPrec1)
+
+mkShowInstance :: DsMonad q => DType -> [DCon] -> q UInstDecl
+mkShowInstance ty cons = do
+  when (null cons) $
```
RyanGlScott: Well, standalone deriving _can_ handle it, but it's perhaps not the implementation you'd expect:

```
λ> data Foo
λ> deriving instance Show Foo

==================== Derived instances ====================
Derived instances:
  instance GHC.Show.Show Ghci4.Foo where
    GHC.Show.showsPrec = GHC.Err.error "Void showsPrec"
```

We don't have the ability to promote `error` yet, so that's not an option. And I'm not sure what other implementation you have in mind for empty datatypes, so I left it as an error.

src/Data/Singletons/Deriving/Show.hs
```diff
@@ -0,0 +1,140 @@
...
+    fail ("Can't derive Show instance for " ++ pprint (typeToTH ty) ++ ".")
+  clauses <- mk_showsPrec cons
+  return $ InstDecl { id_cxt = inferConstraints (DConPr showName) cons
+                    , id_name = showName
+                    , id_arg_tys = [ty]
+                    , id_meths = [ (showsPrecName, UFunction clauses) ] }
+
+mk_showsPrec :: DsMonad q => [DCon] -> q [DClause]
+mk_showsPrec cons = do
+    p <- newUniqueName "p" -- The precedence argument (not always used)
```
RyanGlScott: In the code for `mk_showsPrec_clause`, I use `DWildPa`s in place of `DVarPa p` whenever `p` isn't used, so there's no need to prefix it with an underscore.

src/Data/Singletons/Deriving/Show.hs
RyanGlScott: Ah yes. I thought that GHC turned `data Foo = Foo {}` into a `NormalC`, but that's only if you _reify_ it. `[d| data Foo = Foo {} |]` does get turned into `RecC`, as I just experimentally confirmed. I'll edit the comment accordingly.

src/Data/Singletons/Prelude.hs
RyanGlScott: Ditto my response.

src/Data/Singletons/Prelude/Show.hs
RyanGlScott: I'll do that.

src/Data/Singletons/Prelude/Show.hs
RyanGlScott: That's a good point. What we should _really_ be doing is defining a `Monoid` instance for `Symbols` first, and then use `Data.Monoid.<>` instead of defining it here. But we don't have a `Monoid` class in `singletons` yet, so I went with the simpler approach of rolling my own.

src/Data/Singletons/Prelude/Show.hs
RyanGlScott: I could rename `show'` to `show_`. But that presents a problem when actually defining. Defining `maybe_` is simple, since you can just use `singletons`:

```haskell

$(singletons [d|
  -- Renamed to avoid name clash
  maybe_ :: b -> (a -> b) -> Maybe a -> b
  maybe_ n _ Nothing  = n
  maybe_ _ f (Just x) = f x
 |])
```

But we can't do the same for `show_`, since it's a class method of `Show`, which we need to define with `singletonsOnly`, not `singletons`, as we don't want to define another `Show` class. Would you be alright with me just defining this locally?

```haskell
show_ :: Show a => a -> String
show_ = Prelude.show
```

src/Data/Singletons/Prelude/Show.hs
RyanGlScott: I was afraid you'd ask that :)

I suppose I could puzzle my way to an implementation, if you'd want that. I opted not to in this iteration since the [value-level `Integer -> String`](https://github.com/ghc/ghc/blob/ba597c1dd1daf9643b72dc7aeace8d6b3fce84eb/libraries/base/GHC/Show.hs#L471-L548) function that GHC implements is too complicated to promote, and I'm not clever enough to figure out how to make a simpler version off the top of my head.

src/Data/Singletons/Prelude/Show.hs
RyanGlScott: I manually defined a `Show [a]` instance above. I didn't derive it since lists get shown specially.

src/Data/Promotion/Prelude.hs
RyanGlScott: A consequence of this, unfortunately, is that we have to add more `import Data.Singletons.Prelude.Show` statements to various test cases to work around #190. Oh well.

src/Data/Singletons/Prelude/Show.hs
goldfirere: Yes, that's what I meant. Thanks.

goldfirere: Good grief that's complex. I was just thinking of
```
-- works only for non-negatives
toString :: Integer -> String
toString 0 = "0"
toString 1 = "1"
toString 2 = "2"
...
toString 9 = "9"
toString n = toString (n `div` 10) <> toString (n `mod` 10)
```

But you don't have to if that runs into any trouble.

goldfirere: Oops. Commented in the wrong spot, clearly.

src/Data/Singletons/Prelude/Show.hs
goldfirere: Good grief that's complex. I was just thinking of
```
-- works only for non-negatives
toString :: Integer -> String
toString 0 = "0"
toString 1 = "1"
toString 2 = "2"
...
toString 9 = "9"
toString n = toString (n `div` 10) <> toString (n `mod` 10)
```

But you don't have to if that runs into any trouble.

RyanGlScott: How would you promote `div` and `mod`? Both of them also rely on primitive `integer-gmp` operations.

goldfirere: See, for example, https://coq.inria.fr/library/Coq.Init.Nat.html

vagarenko: Why `Div` and `Mod` are not implemented in GHC.TypeLits?
I asked some time ago on trac https://ghc.haskell.org/trac/ghc/ticket/13652 , but no one answered.

RyanGlScott: @vagarenko, the answer to the question "why isn't this in `base`?" is almost always "because you haven't implemented it yet", and that holds true here as well.

That being said, Trac is not the place to make requests to add things to the standard libraries. That's better suited to a discussion on the [libraries mailing list](https://mail.haskell.org/mailman/listinfo/libraries).

RyanGlScott:
I've added a dog-slow `PShow Nat` instance, per your request. As @vagarenko, it could be a lot more efficient if we had built-in `Divmod`, `Div`, and `Mod` type families in `GHC.TypeNats`, but that's not an option at the moment.

I believe I've addressed all of your comments. Let me know if you'd like any other changes.

goldfirere:
Thanks -- merged.

Issue #192, opened 2017-06-07
Deriving Show fails to discover local fixity declaration


RyanGlScott:
This file should typecheck, but doesn't:

```haskell
{-# LANGUAGE ExistentialQuantification #-}
{-# LANGUAGE InstanceSigs #-}
{-# LANGUAGE ScopedTypeVariables #-}
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE TypeInType #-}
{-# LANGUAGE TypeOperators #-}
{-# LANGUAGE UndecidableInstances #-}
module Bug where

import Data.Singletons.Prelude
import Data.Singletons.TH
import Data.Type.Equality

$(singletons [d|

  infixl 6 :*:
  data T = () :*: () deriving Show

  |])

test1 :: "() :*: ()" :~: ShowsPrec 6 ('() :*: '()) ""
test1 = Refl

test2 :: "(() :*: ())" :~: ShowsPrec 7 ('() :*: '()) ""
test2 = Refl
```

```
GHCi, version 8.2.0.20170522: http://www.haskell.org/ghc/  :? for help
Loaded GHCi configuration from /home/rgscott/.ghci
[1 of 1] Compiling Bug              ( Bug.hs, interpreted )

Bug.hs:26:9: error:
    • Couldn't match type ‘"(() :*: ())"’ with ‘"() :*: ()"’
      Expected type: "(() :*: ())" :~: ShowsPrec 7 ('() ':*: '()) ""
        Actual type: "(() :*: ())" :~: "(() :*: ())"
    • In the expression: Refl
      In an equation for ‘test2’: test2 = Refl
   |
26 | test2 = Refl
   |         ^^^^
```

In plain terms, since the fixity of `(:*:)` is 6, then showing it with a precedence of 6 should not surround the output with parentheses (which is working), but showing it with a precedence of 7 _should_ surround the output with parentheses (which is not working).

The culprit is that `deriving Show` is failing to discover the local fixity declaration for `(:*:)`. This can be seen if you inspect the `-ddump-splices` output:

```haskell
type family ShowsPrec_6 (p :: Nat) (x :: T) (s :: Symbol) :: Symbol where
[...]

PR #194, opened 2017-06-07
Add support for standalone deriving declarations

RyanGlScott:
This adds the ability to promote and single standalone `deriving` declarations. This provides a usable workaround for #192.

### DIFF
--- a/src/Data/Singletons/Partition.hs
+++ b/src/Data/Singletons/Partition.hs
@@ -37,12 +37,13 @@ data PartitionedDecs =
         , pd_class_decs :: [UClassDecl]
         , pd_instance_decs :: [UInstDecl]
         , pd_data_decs :: [DataDecl]
+        , pd_standalone_derived_eq_decs :: [StandaloneDerivedEqDec]
         }
 
 instance Monoid PartitionedDecs where
-  mempty = PDecs [] [] [] []
-  mappend (PDecs a1 b1 c1 d1) (PDecs a2 b2 c2 d2) =
-    PDecs (a1 <> a2) (b1 <> b2) (c1 <> c2) (d1 <> d2)
+  mempty = PDecs [] [] [] [] []
+  mappend (PDecs a1 b1 c1 d1 e1) (PDecs a2 b2 c2 d2 e2) =
+    PDecs (a1 <> a2) (b1 <> b2) (c1 <> c2) (d1 <> d2) (e1 <> e2)
 
 -- | Split up a @[DDec]@ into its pieces, extracting 'Ord' instances
 -- from deriving clauses
@@ -103,7 +104,15 @@ partitionDec (DStandaloneDerivD mb_strat ctxt ty) =
               Just (DTyConI (DDataD _ _ _ _ cons _) _) -> do
                 mb_instance <- partitionDeriving mb_strat cls_pred (Just ctxt) data_ty cons
                 case mb_instance of
-                  Left _ -> return mempty -- singletons doesn't support deriving this instance
+                  Left _ -> case cls_pred of
+                              -- TODO: Note
+                              DConPr cls_name
+                                | isStock mb_strat, cls_name == eqName
+                                -> let sded = SDEqDec { sded_cxt  = ctxt
+                                                      , sded_type = data_ty
+                                                      , sded_cons = cons }
+                                   in return mempty { pd_standalone_derived_eq_decs = [sded] }
+                              _ -> return mempty -- singletons doesn't support deriving this instance
                   Right derived_instance -> return $ mempty { pd_instance_decs = [derived_instance] }
               Just _ ->
                 fail $ "Standalone derived instance for something other than a datatype: "
@@ -147,8 +156,14 @@ partitionDeriving mb_strat deriv_pred mb_ctxt ty cons = case deriv_pred of
      | stock, deriv_name == showName
     -> Right <$> mkShowInstance mb_ctxt ty cons
     where
-      stock = case mb_strat of
-                Nothing            -> True
-                Just StockStrategy -> True
-                Just _             -> False
+      stock = isStock mb_strat
   _ -> return (Left deriv_pred)
+
+isStock :: Maybe DerivStrategy -> Bool
+isStock Nothing = True -- We assume the lack of an explicit deriving strategy to
+                       -- indicate defaulting to stock. In reality, GHC's defaulting
+                       -- behavior is much more complex than this, but implementing
+                       -- this in singletons would be impossible, as would require
+                       -- detecting the presence of extensions.
+isStock (Just StockStrategy) = True
+isStock (Just _)             = False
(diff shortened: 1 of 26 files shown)


tests/compile-and-dump/Singletons/StandaloneDeriving.hs
```diff
@@ -0,0 +1,32 @@
...
+
+import Data.Singletons.Prelude
+import Data.Singletons.Prelude.Show
+import Data.Singletons.TH
+
+$(singletons [d|
+
+  infixl 6 :*:
+  data T a = a :*: a
+  data S = S1 | S2 deriving Eq
```
RyanGlScott: I labeled this WIP because there's one awkward point to deal with: `deriving Eq`. You'll note that I didn't standalone derive `Eq` in this example because, well, it would get ignored. That's because for some odd reason, the logic for detecting derived `Eq` instances is sequestered off in [`Data.Singletons.Single.Data`](https://github.com/goldfirere/singletons/blob/75dae57bbe06b21b3adbb82158c4767ebd695d14/src/Data/Singletons/Single/Data.hs#L59-L62), instead of `Data.Singletons.Promote.Partition`, where every other class is detected.

Is there a reason for doing it this way? If so, I need to figure out how to adapt this approach for standalone derived `Eq` instances...

goldfirere: The problem with `Eq` is that deriving `Eq` gets you `PEq` and `SEq`, *and* it gets you `SDecide`. That last one can't be defined in the `Partition` module, because it's not the result of promoting/singling, unlike all the other instances.

RyanGlScott: OK, after reading the come a little more carefully, it appears that we call `partitionDecs` _before_ `singDataD`/`promoteDataDec`, which handle `Eq`. So I believe what I need to do is:

1. Make it so that `partitionDecs` does _not_ discard standalone deriving declarations it cannot handle, but instead defer them so that `singDataD`/`promoteDataDec` may act on them. I believe I could accomplish this by adding another field to [`PartitionedDecs`](https://github.com/goldfirere/singletons/blob/75dae57bbe06b21b3adbb82158c4767ebd695d14/src/Data/Singletons/Partition.hs#L34) for deferred standalone deriving declarations.
2. In `singDataD`/`promoteDataDec`, check the deferred standalone deriving declarations for `Eq`. If it's there, derive `SEq`/`PEq`/`SDecide`.
3. Otherwise, discard the deferred standalone deriving declarations, since `singletons` can't handle them.

Does that sound agreeable?

goldfirere: So `partitionDecs` would still handle standalone-deriving for non-`Eq` classes? If "yes" (which I think is reasonable) then there is no need for the full generality of standalone-deriving declarations in `PartitionedDecs`. Just add a field which is a list of names for which there is a standalone-deriving for `Eq`. Or do we need the instance contexts, too?

RyanGlScott: > So partitionDecs would still handle standalone-deriving for non-Eq classes?

Yes.

> Just add a field which is a list of names for which there is a standalone-deriving for Eq. Or do we need the instance contexts, too?

Yes, you would need the contexts, since you could easy do something like `deriving instance Eq a => Eq (Either a Int)`, but the context you'd get from attaching a `deriving Eq` clause to `data Either a b` would be `(Eq a, Eq b)`. But your point about only needing to keep the bits related to standalone-derived `Eq` instances is well taken. I'll do that (and add a Note, since this is somewhat of an unintuitive design choice).

goldfirere: Sounds good. It is a bit of an odd design, but there is a reason for it!

tests/compile-and-dump/Singletons/StandaloneDeriving.ghc82.template
```diff
@@ -0,0 +1,448 @@
...
+Singletons/StandaloneDeriving.hs:(0,0)-(0,0): Splicing declarations
+    singletons
+      [d| deriving instance Enum S
+          deriving instance Bounded S
+          deriving instance Show S
+          deriving instance Ord S
+          deriving instance Eq S
+          deriving instance Show a => Show (T a ())
+          deriving instance Ord a => Ord (T a ())
+          deriving instance Eq a => Eq (T a ()) |]
```
RyanGlScott: Huh, that's not the order in which I originally wrote those! https://ghc.haskell.org/trac/ghc/ticket/13799

RyanGlScott:
How does this look, @goldfirere? Any changes that should be made?

(I don't have push access to `singletons`, so I'm stalled for now.)

src/Data/Singletons/Partition.hs
```diff
@@ -128,3 +142,28 @@ partitionInstanceDec (DLetDec (DFunD name clauses)) =
 partitionInstanceDec (DLetDec (DPragmaD {})) = return Nothing
 partitionInstanceDec _ =
   fail "Only method bodies can be promoted within an instance."
+
+partitionDeriving :: DsMonad m => Maybe DerivStrategy -> DPred -> Maybe DCxt -> DType -> [DCon]
```
goldfirere: It seems that this should really return one of *three* possibilities: an instance, a `SDEqDec`, or neither of the above. Perhaps the best way for it to do this is for the function to return a `PartitionedDecs` that can be `<>`ed with other declarations as necessary. That seems to be a cleaner design. (Then, code elsewhere need not detect `Eq` in a `deriving` clause.)

src/Data/Singletons/Single.hs
```diff
@@ -555,6 +558,40 @@ singExp (ADLetE env exp) res_ki =
...
+  (scons, _) <- singM [] $ mapM (singCtor aVar) cons
+  sctxt <- mapM singPred ctxt
+  sEqInst <- mkEqualityInstance (Just sctxt) ty scons sEqClassDesc
+  -- Beware! The user might have specified an instance context like this:
+  --
+  --   deriving instance Eq a => Eq (T a Int)
+  --
+  -- When we single the context, it will become (SEq a). But we do *not* want
+  -- this for the SDecide instance! The simplest solution is to simply replace
+  -- all occurrences of SEq with SDecide in the context.
```
goldfirere: Agreed.

src/Data/Singletons/Syntax.hs
```diff
@@ -132,3 +132,29 @@ buildLetDecEnv = go emptyLetDecEnv
     go acc (DInfixD f n : rest) =
       go (infixDecl f n <> acc) rest
     go acc (DPragmaD{} : rest) = go acc rest
+
+-- See Note [Standalone derived Eq instances]
+data StandaloneDerivedEqDec = SDEqDec
```
goldfirere: Under my proposed refactoring, this would handle all `Eq` instances, not just standalone ones. It could then hold a `Maybe DCxt` instead of a `DCxt`.

src/Data/Singletons/Partition.hs
RyanGlScott: Agreed! I've pushed this change.

goldfirere:
The revisions look good. Please rebase and merge. Thanks!

PR #196, opened 2017-06-13
Fix #192 by increasing the scope of withLocalDeclarations

### LINKED ISSUE #192: Deriving Show fails to discover local fixity declaration

RyanGlScott:
Previously, we were calling `mkShowInstance` without a surrounding `withLocalDeclarations` to bring in locally declared fixity declarations, which led to incorrect fixity reification. I've fixed this by making `withLocalDeclarations` have a wider berth.

### DIFF
--- a/src/Data/Singletons/Single.hs
+++ b/src/Data/Singletons/Single.hs
@@ -202,8 +202,8 @@ singInfo (DPatSynI {}) =
   fail "Singling of pattern synonym info not supported"
 
 singTopLevelDecs :: DsMonad q => [Dec] -> [DDec] -> q [DDec]
-singTopLevelDecs locals raw_decls = do
-  decls <- withLocalDeclarations locals $ expand raw_decls     -- expand type synonyms
+singTopLevelDecs locals raw_decls = withLocalDeclarations locals $ do
+  decls <- expand raw_decls     -- expand type synonyms
   PDecs { pd_let_decs              = letDecls
         , pd_class_decs            = classes
         , pd_instance_decs         = insts
--- a/tests/compile-and-dump/Singletons/ShowDeriving.ghc82.template
+++ b/tests/compile-and-dump/Singletons/ShowDeriving.ghc82.template
@@ -1,6 +1,10 @@
 Singletons/ShowDeriving.hs:(0,0)-(0,0): Splicing declarations
     singletons
-      [d| data Foo1
+      [d| infixl 5 `MkFoo2b`
+          infixl 5 :*:
+          infixl 5 :&:
+          
+          data Foo1
             = MkFoo1
             deriving Show
           data Foo2 a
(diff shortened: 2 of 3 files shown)


goldfirere:
Oh yes. That was very silly of me.

Issue #197, opened 2017-06-14
"Multiple fixity declarations" errors


RyanGlScott:
I discovered this when investigating #160. This is reproducble with `singletons-2.3` or HEAD:

```haskell
{-# LANGUAGE ExistentialQuantification #-}
{-# LANGUAGE ScopedTypeVariables #-}
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE TypeInType #-}
{-# LANGUAGE TypeOperators #-}
{-# LANGUAGE UndecidableInstances #-}
module Bug where

import Data.Singletons.TH

$(singletons [d|
  infixl 5 $$:
  ($$:) :: Bool -> Bool -> Bool
  _ $$: _ = False
 |])
```
```
$ /opt/ghc/8.2.1/bin/runghc -ddump-splices Bug.hs
Bug.hs:(12,3)-(16,3): Splicing declarations
    singletons
      [d| infixl 5 $$:_a3AU
          
          ($$:_a3AU) :: Bool -> Bool -> Bool
          _ $$:_a3AU _ = False |]
  ======>
    infixl 5 $$:_a6jG
    ($$:_a6jG) :: Bool -> Bool -> Bool
    ($$:_a6jG) _ _ = False
    type ($$:$$$) (t_a6jH :: Bool) (t_a6jI :: Bool) =
        ($$:) t_a6jH t_a6jI
    instance SuppressUnusedWarnings ($$:$$) where
      suppressUnusedWarnings _
        = snd ((GHC.Tuple.(,) (:$$:$$###)) GHC.Tuple.())
    data ($$:$$) (l_a6jK :: Bool) (l_a6jJ :: TyFun Bool Bool)
      = forall arg_a6jL. SameKind (Apply (($$:$$) l_a6jK) arg_a6jL) (($$:$$$) l_a6jK arg_a6jL) =>
        (:$$:$$###)
    type instance Apply (($$:$$) l_a6jK) l_a6jJ = ($$:) l_a6jK l_a6jJ
    instance SuppressUnusedWarnings ($$:$) where
      suppressUnusedWarnings _
        = snd ((GHC.Tuple.(,) (:$$:$###)) GHC.Tuple.())
    data ($$:$) (l_a6jM :: TyFun Bool (TyFun Bool Bool
                                       -> GHC.Types.Type))
[...]

PR #199, opened 2017-06-14
Add special casing for $ when promoting fixity declarations

RyanGlScott:
We needn't promote fixity declarations for identifiers that begin with `$`, since their promoted names are the same as their value-level names. This addresses the first part of #197.


goldfirere:
Although I think we should find a different solution than special-casing `$`, the current state of affairs (with a partial special-case for `$`) is broken, and this fixes it. Thanks.

Issue #172, opened 2016-12-09
Illegal data constructor name: ‘$>$$###’


RyanGlScott:
```haskell
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE ExistentialQuantification #-}
{-# LANGUAGE PolyKinds #-}
{-# LANGUAGE ScopedTypeVariables #-}
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE TypeOperators #-}
{-# LANGUAGE UndecidableInstances #-}
module Bug where

import Data.Singletons.Prelude
import Data.Singletons.TH
import Data.Singletons.TypeLits

$(singletonsOnly [d|
  ($>) :: Nat -> Nat -> Nat
  ($>) = (+)
  |])
```

```
$ /opt/ghc/8.0.1/bin/ghc Bug.hs 
[1 of 1] Compiling Bug              ( Bug.hs, Bug.o )

Bug.hs:15:3: error:
    Illegal data constructor name: ‘$>$$###’
    When splicing a TH declaration:
      instance Data.Singletons.SuppressUnusedWarnings.SuppressUnusedWarnings $>$$
    where Data.Singletons.SuppressUnusedWarnings.suppressUnusedWarnings _ = Data.Tuple.snd (GHC.Tuple.(,) ($>$$###) GHC.Tuple.())
```

This only seems to affect infix functions that begin with `$`. In contrast, trying to promote, say, `<$>` works without issue.

RyanGlScott:
Thanks! But I suppose we should make mention of this special-casing [here](https://github.com/goldfirere/singletons/tree/06d1d6879cb06bdb419b2b69cf27cedf4b5d1790#on-names)?

goldfirere:
Not necessary, as the only name that changed was for the "kind inference" data constructor, which is purely internal. Thanks for thinking of this, though!

Issue #29, opened 2014-04-23
Can't promote ($) because of our naming conventions


jstolarek:
Turns out we can't promote `($)` application operator because of our naming conventions. Reason: we prepend it with `:` and now the name collides with symbols for `:` (cons), which is appended with `$`. I tried to create some ad-hoc solution to this but gave up after reaching errors with incorrect name of  `$$$###` constructor. At the moment I don't know what are these `###` things.

PR #203, opened 2017-06-17
Eliminate the dollar sign hack

RyanGlScott:
`singletons` currently special-cases a _lot_ of operations for names that begin with `$`. It has been the center of a spider's nest of bugs (see #29, https://github.com/goldfirere/singletons/issues/160#issuecomment-308519043, #172, and #197 for examples).

This PR rips out the hack, and improves the situation surrounding defunctionalization symbols, which use `$` characters as suffixes. We now precede the defunctionalized `$`s with the string `@#$%^%$#@` (as suggested in https://github.com/goldfirere/singletons/issues/197#issuecomment-308609773). This allows use to promote `($)` and `($!)` without any special-casing at all. Huzzah!


README.md
```diff
@@ -433,7 +433,7 @@ There are some special cases:
 
    singleton constructr: `SCons`
 
-   symbols: `ConsSym0`, `ConsSym1`
+   symbols: `:@#$%^%$#@$`, `:@#$%^%$#@$`
```
RyanGlScott: Note: I didn't change conventions here! I checked, and it turns out the defunctionalization symbols for `:` weren't actually named `ConsSym0` and `ConsSym1` at all, but rather `(:$)` and `(:$$)`. I just opted to update this reference to be correct here, since I happened to change what `(:$)` and `(:$$)` were named.

RyanGlScott:
The downside to this is that now defunctionalization symbols are obscenely long and hard to remember. Perhaps we ought to provide a convenience function of type `Name -> Q Type` which gives you the appropriate defunctionalization symbol on demand?

goldfirere:
This diff is *ugly*. Is there an intermediate solution, something without quite so many symbols? Like, maybe 2-3 arbitrary symbols will be enough in practice. (Note that if someone wrote a normal identifier that happened to end in `Sym0`, we'd hit a similar problem with alpha identifiers.) In any case, you're free to merge this, but if it's easy to shorten the glyph string, that would seem to be an improvement.

RyanGlScott:
Indeed, I'm not claiming this will eliminate naming conflicts altogether (ultimately, we'd need something like #204 to work around that robustly), but it will certainly make it less likely.

How about `@#@` instead for the suffix?

goldfirere:
`@#@` is great. Thanks!

PR #205, opened 2017-06-19
Make the treatment of undefined less magical

### LINKED ISSUE #202: Revisit the special case for undefined
Currently, we [special-case the promotion of `undefined`](https://github.com/goldfirere/singletons/tree/c1561a855c71739434962784bc6fb3c2b1a8b785#special-names):

> 6. original value: `undefined`
>     promoted type: `Any`
>     singleton value: `undefined`
>     symbols: Any

But do we need to? After all, these days it's perfectly possible to define `Any` oneself:

```haskell
type family Any :: k where {}
```

Couldn't we just as easily call this `Undefined`?

An added bonus is that this would let us avoid the name `any_` for the list operation.

RyanGlScott:
This lets us remove some special-casing, which is always nice. Note that we can't remote _all_ special casing (see `isException`, for instance), but it's a fair deal better than before.

Fixes #202.

### DIFF
--- a/src/Data/Singletons/Single.hs
+++ b/src/Data/Singletons/Single.hs
@@ -590,7 +590,7 @@ sEqToSDecide p@(DConPr n)
 sEqToSDecide DWildCardPr = DWildCardPr
 
 isException :: DExp -> Bool
-isException (DVarE n)             = n == undefinedName
+isException (DVarE n)             = nameBase n == "sUndefined"
 isException (DConE {})            = False
 isException (DLitE {})            = False
 isException (DAppE (DVarE fun) _) | nameBase fun == "sError" = True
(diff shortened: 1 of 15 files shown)


src/Data/Singletons/Single.hs
```diff
@@ -590,7 +590,7 @@ sEqToSDecide p@(DConPr n)
 sEqToSDecide DWildCardPr = DWildCardPr
 
 isException :: DExp -> Bool
-isException (DVarE n)             = n == undefinedName
+isException (DVarE n)             = nameBase n == "sUndefined"
```
goldfirere: Can we do better here? i.e. by wiring in `sUndefined` in `Names`.

RyanGlScott: By "do better", do you mean check if `nameBase n == nameBase 'sUndefined`? I don't think we can check if it's directly equal to ``sUndefined`, since I believe we're checking a `Name` which has been stubbed out by Template Haskell, so it won't have the right namespace.

goldfirere: Ah. Right you are. Never mind then.

Issue #219, opened 2017-07-08
Definition using foldr1 doesn't singletonize


RyanGlScott:
I was surprised to discover this when attempting to implement `unwords` (#212). Here's something similar, which doesn't use `Symbol` (so that it can be tested on both `singletons-2.3` and HEAD):

```haskell
{-# LANGUAGE ExistentialQuantification #-}
{-# LANGUAGE ScopedTypeVariables #-}
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE TypeInType #-}
{-# LANGUAGE UndecidableInstances #-}
module Bug where

import Data.Singletons.Prelude
import Data.Singletons.TH

$(singletons [d|
    f :: [[a]] -> [a]
    f [] = []
    f ws = foldr1 (++) ws
  |])
```

```
../Bug.hs:(12,3)-(16,4): Splicing declarations
    singletons
      [d| f_a3l25 :: [[a_a3l26]] -> [a_a3l26]
          f_a3l25 [] = []
          f_a3l25 ws_a3l27 = foldr1 (++) ws_a3l27 |]
  ======>
    f_a3l33 :: [[a_a3l32]] -> [a_a3l32]
    f_a3l33 ghc-prim-0.5.0.0:GHC.Types.[] = []
    f_a3l33 ws_a3l34 = (foldr1 (++)) ws_a3l34
    type FSym1 (t_a3l35 :: [[a6989586621679805688]]) = F t_a3l35
    instance SuppressUnusedWarnings FSym0 where
      suppressUnusedWarnings
        = snd
            ((ghc-prim-0.5.0.0:GHC.Tuple.(,) FSym0KindInference)
               ghc-prim-0.5.0.0:GHC.Tuple.())
    data FSym0 (l_a3l36 :: TyFun [[a6989586621679805688]] [a6989586621679805688])
      = forall arg_a3l37. SameKind (Apply FSym0 arg_a3l37) (FSym1 arg_a3l37) =>
        FSym0KindInference
    type instance Apply FSym0 l_a3l36 = F l_a3l36
    type family F (a_a3l38 :: [[a_a3l32]]) :: [a_a3l32] where
      F '[] = '[]
      F ws_a3l39 = Apply (Apply Foldr1Sym0 (:++@#@$)) ws_a3l39
    sF ::
      forall (t_a3l3a :: [[a_a3l32]]).
[...]

PR #220, opened 2017-07-08
Add unwords and unlines

### LINKED ISSUE #212: Promote unwords and unlines
Currently, `lines`, `words`, `unwords`, and `unlines` are [commented out](https://github.com/goldfirere/singletons/blob/ed0dc6c1e0fd0aabf513df8b4a419529fe4678b4/src/Data/Singletons/Prelude/List.hs#L497-L521) in `Data.Singletons.Prelude.List`, and preceded with this comment:

```
-- We can't promote any of these functions because at the type level
-- String literals are no longer considered to be lists of Chars, so
-- there is mismatch between term-level and type-level semantics
```

This certainly holds true for `lines` and `words`. But now that we have type-level `Symbol` appending, it's quite possibly to promote `unwords` and `unlines` in GHC 8.2:

```haskell
unlines                 :: [Symbol] -> Symbol
unlines                 =  concatMap (<> "\n")

unwords                 :: [Symbol] -> Symbol
unwords []              =  ""
unwords ws              =  foldr1 (\w s -> w <> " " <> s) ws
```

The only problem is that trying to add this directly would lead to an import cycle between `Data.Singletons.Prelude.List` and `Data.Singletons.Prelude.Show`. Urk. I suppose it's time to move some definitions from one module or the other to an `Internal` module.

RyanGlScott:
I broke the import cycle between `Data.Singletons.Prelude.List` and `Data.Singletons.Prelude.Show` by migrating `:<>` to `Data.Singletons.Prelude.TypeLits`, which is probably a better home for it anyways.

Note that I had to use an alternative implementation of `unwords`, since the one from the `Prelude` doesn't singletonize due to #219.

Fixes #212.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,11 @@ next
   corresponding promoted type, singleton function, and defunctionalization
   symbols are now named `Any`, `sAny`, and `AnySym{0,1,2}`.
 
+* Add `(:<>)` and `(%:<>)`, the promoted and singled versions of `AppendSymbol`
+  from `GHC.TypeLits`.
+
+* Add `unlines` and `unwords` to `Data.Singletons.Prelude.List`.
+
 * Add promoted and singled versions of `Show`, including `deriving` support.
 
 * Permit derived `Ord` instances for empty datatypes.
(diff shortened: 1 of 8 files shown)


goldfirere:
This change is forward compatible with a promoted `Monoid` instance, isn't it? I do believe it is. Then I'm very happy with this addition. Thanks!

PR #221, opened 2017-07-09
Rework handling of empty data types w.r.t. SingKind, PEq, SEq, and SDecide

### LINKED ISSUE #218: Derived Eq and Ord instances for empty data types contradict each other
While attempting to implement #217, I noticed something strange:

```
λ> ; $(singletons [d| data Void; deriving instance Eq Void; deriving instance Ord Void |])
<interactive>:25:5-86: Splicing declarations
    singletons
      [d| data Void_aahx
          
          deriving instance Ord Void_aahx
          deriving instance Eq Void_aahx |]
  ======>
    ...
    type family Compare_6989586621679049364_aaih (a_aaie :: Void_aai4) (a_aaif :: Void_aai4) :: Ordering where
      Compare_6989586621679049364_aaih _ _ = EQSym0
    ...
    type family Equals_6989586621679049373_aaiq (a_aair :: Void_aai4) (b_aais :: Void_aai4) :: Bool where
      Equals_6989586621679049373_aaiq (a_aait :: Void_aai4) (b_aaiu :: Void_aai4) = FalseSym0
```

Note that the `Eq` and `Ord` instances don't agree with each other! The `Eq` instance (via `Equals`) claims that no two types of kind `Void` are equal, whereas the `Ord` instance (via `Compare`) believes that //any// two types of kind `Void` are equal.

For consistency, I think the derived `Eq` instance should always return `True` in the case of empty data types.

RyanGlScott:
This is an overhaul of the way `singletons` handles empty data types (e.g., `data Void`). There are four main changes:

* Before, derived `PEq` instances for empty data types would always return `False`. I've changed this to return `True` instead, for consistency with derived `POrd` instances.
* Before, derived `SEq` instances for empty data types would always error. I've changed this to return `STrue` instead for consistency with derived `PEq` instances.
* Before, derived `SDecide` instances for empty data types would always error. I've changed this to return `Proved ⊥` instead, for (you guessed it) consistency.
* Before, generated `SingKind` instances for empty data types would always error when `fromSing` or `toSing` were invoked. I've changed this so that `fromSing x = case x of {}` and `toSing x = SomeSing (case x of {})`.

The downside is that you'll need to enable `EmptyCase` more often, but given that this has been around since GHC 7.8 and is a pretty benign extension, I think it's pretty reasonable to require.

Fixes #218.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,14 @@ next
   corresponding promoted type, singleton function, and defunctionalization
   symbols are now named `Any`, `sAny`, and `AnySym{0,1,2}`.
 
+* Rework the treatment of empty data types:
+  * Generated `SingKind` instances for empty data types now use `EmptyCase`
+    instead of simply `error`ing.
+  * Derived `PEq` instances for empty data types now return `True` instead of
+    `False`. Derived `SEq` instances now return `True` instead of `error`ing.
+  * Derived `SDecide` instances for empty data types now return `Proved bottom`,
+    where `bottom` is a divergent computation, instead of `error`ing.
+
 * Add `(:<>)` and `(%:<>)`, the promoted and singled versions of `AppendSymbol`
   from `GHC.TypeLits`.
 
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ extensions. This list includes, but is not necessarily limited to, the
 following:
 
 * `DefaultSignatures`
+* `EmptyCase`
 * `ExistentialQuantification`
 * `FlexibleContexts`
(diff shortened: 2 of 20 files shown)


int-index:
> a pretty benign extension

The issue with `EmptyCase` is that the compiler doesn't even try to ascertain that there's indeed no cases.

RyanGlScott:
Actually, as of GHC 8.2, there is indeed exhaustiveness checking for `EmptyCase`! See  https://ghc.haskell.org/trac/ghc/ticket/10746.

PR #224, opened 2017-07-10
Add Eq, Ord, Num, Enum, and Bounded instances for SomeSing

### LINKED ISSUE #215: Eq and Ord instances for SomeSing
In the vein of @andrewthad's [`pringletons`](http://hackage.haskell.org/package/pringletons) library, I'd like to write `Eq` and `Ord` instances for `SomeSing`:

```haskell
instance SEq k => Eq (SomeSing k) where
  SomeSing a == SomeSing b = fromSing (a %:== b)

instance SOrd k => Ord (SomeSing k) where
  SomeSing a `compare` SomeSing b = fromSing (a `sCompare` b)
```

This is useful, for instance, for storing `SomeSing`s in `Map`s.

There's one problem with this idea: it's difficult to add these instances as non-orphans. We can't simply add them to `Data.Singletons`, as that would create import cycles with `Data.Singletons.Prelude.Eq` and `Data.Singletons.Prelude.Ord`. One solution is to add `Data.Singletons.Prelude.Eq.Internal` and `Data.Singletons.Prelude.Ord.Internal` modules to break the import cycles (which is something that probably needs to be done anyway, since a similar problem arose in https://github.com/goldfirere/singletons/issues/212).

RyanGlScott:
To accomplish this, I created a new `Data.Singletons.Internal` module to avoid the import cycle that would otherwise ensue when you combine `SomeSing` with `SEq`, `SOrd`, etc.

Fixes #215.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -41,6 +41,10 @@ next
 
 * Add a value-level `(@@)`, which is a synonym for `applySing`.
 
+* Add `Eq`, `Ord`, `Num`, `Enum`, and `Bounded` instances for `SomeSing`, which
+  leverage the `SEq`, `SOrd`, `SNum`, `SEnum`, and `SBounded` instances,
+  respectively, for the underlying `Sing`.
+
 2.3
 ---
 * Documentation clarifiation in `Data.Singletons.TypeLits`, thanks to @ivan-m.
--- a/singletons.cabal
+++ b/singletons.cabal
@@ -97,6 +97,7 @@ library
                       Data.Singletons.Deriving.Enum,
                       Data.Singletons.Deriving.Ord,
                       Data.Singletons.Deriving.Show,
+                      Data.Singletons.Internal,
                       Data.Singletons.Prelude.List.NonEmpty.Internal,
                       Data.Singletons.Promote,
                       Data.Singletons.Promote.Monad,
--- a/src/Data/Promotion/TH.hs
+++ b/src/Data/Promotion/TH.hs
@@ -79,7 +79,7 @@ module Data.Promotion.TH (
 
  ) where
 
-import Data.Singletons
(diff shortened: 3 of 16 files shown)


goldfirere:
I suppose this design makes sense. I've always liked that `Data.Singletons` had no dependencies on any of the TH stuff, but I suppose that matters only for developers of `singletons`. Users will have all the other stuff anyway.

I *have* wondered if `singletons` should be broken into three packages: the basic definitions (the old `Data.Singletons` module), `singletons-th` which defines the TH support, and `singletons-base`, which provides the "standard library". I'm not currently advocating for such a change, but just sharing the idea in case others have thought similarly.

RyanGlScott:
In practice, I don't know how feasible it would be to separate `singletons-base` from `singletons-th`. The problem is that a sizable chunk of the standard library is itself generated by Template Haskell! Splitting up `singletons` just to avoid incurring a TH dependency feels rather pointless to me, especially since there are folks working hard to make sure TH works everywhere soon.

That being said, I can certainly understand the desire to split off `Data.Singletons` from the rest of this (very heavy) library, and I realize that this PR gets in the way of that, since realizing that change would necessarily require these `SomeSing` instances to be orphans (whereas if everything is is one package like it is now, we can minimize the damage orphan instances do through internal modules). I don't know a good way to have both things.

RyanGlScott:
Actually, there'd also be more orphans if you split off `Data.Singletons` from `singletons`, since the bulk of the `Sing` instances are currently generated via Template Haskell, and I can't say that I'm especially eager to decouple `Sing` from its instances further.

goldfirere:
Go ahead and merge this. My earlier comment was just to describe a path not taken. I don't think we should try to do it, though.

Issue #226, opened 2017-07-10
Promoting infix class names doesn't actually work


RyanGlScott:
The `README` [claims](https://github.com/goldfirere/singletons/tree/f0bc6ebd6d8a3c4a60c9c7d6d605f8d453688225#on-names) that this is how infix class names are promoted/singled:

> 8. original class: `~>`
>     promoted class: `#~>`
>     singleton class: `:%~>`

But that doesn't seem to work if you actually try it out:

```haskell
{-# LANGUAGE MultiParamTypeClasses #-}
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE TypeOperators #-}
module Bug where

import Data.Singletons.TH

$(singletons [d| class a ~> b |])
```
```
../Bug.hs:8:3: error:
    Illegal type constructor or class name: ‘P~>’
    When splicing a TH declaration: class P~> (a_0 :: *) (b_1 :: *)
  |
8 | $(singletons [d| class a ~> b |])
  |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

RyanGlScott:
Ugh, this is more hairiness emerging from `singletons` inability to reason about infix type names that don't begin with a colon. For comparison, this _does_ work:

```haskell
{-# LANGUAGE MultiParamTypeClasses #-}
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE TypeOperators #-}
module Bug where

import Data.Singletons.TH

$(singletons [d| class a :~> b |])
```
```
../Bug.hs:8:3-33: Splicing declarations
    singletons [d| class a_a5WTf :~>_a5WTe b_a5WTg |]
  ======>
    class (:~>_a5WTF) a_a5WTG b_a5WTH
    class (#~>) (a_a5WTG :: ghc-prim-0.5.0.0:GHC.Types.Type) (b_a5WTH :: ghc-prim-0.5.0.0:GHC.Types.Type)
    class (:%~>) a_a5WTG b_a5WTH
```

Since this is basically a dupe of #163, I'll close this.

Issue #200, opened 2017-06-14
Multiple declarations of ‘:$$:$###’


RyanGlScott:
Yet another bug discovered when working on #160. This is only reproducible on `singletons-2.3` or later, since it relies on the fix to #172. This program:

```haskell
{-# LANGUAGE ExistentialQuantification #-}
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE TypeInType #-}
{-# LANGUAGE TypeOperators #-}
module Bug where

import Data.Singletons.TH

$(singletons [d|
      data ErrorMessage = ErrorMessage :$$: ErrorMessage

      ($$:) :: ErrorMessage -> ErrorMessage -> ErrorMessage
      x $$: y = x :$$: y
    |])
```
```
GHCi, version 8.2.0.20170523: http://www.haskell.org/ghc/  :? for help
Loaded GHCi configuration from /home/rgscott/.ghci
[1 of 1] Compiling Bug              ( Bug.hs, interpreted )

Bug.hs:10:3: error:
    Multiple declarations of ‘:$$:$###’
    Declared at: Bug.hs:10:3
                 Bug.hs:10:3
   |
10 | $(singletons [d|
   |   ^^^^^^^^^^^^^^...

Bug.hs:10:3: error:
    Multiple declarations of ‘:$$:$$###’
    Declared at: Bug.hs:10:3
                 Bug.hs:10:3
   |
10 | $(singletons [d|
   |   ^^^^^^^^^^^^^^...
```

This can be seen by inspecting the `-ddump-splices` output:

```
-- For (:$$:)
data (:$$:$$) (l_a6nl :: ErrorMessage_a6ne) (l_a6nk :: TyFun ErrorMessage_a6ne ErrorMessage_a6ne)
  = forall arg_a6nm. SameKind (Apply ((:$$:$$) l_a6nl) arg_a6nm) ((:$$:$$$) l_a6nl arg_a6nm) =>
    (:$$:$$###)
data (:$$:$) (l_a6nn :: TyFun ErrorMessage_a6ne (TyFun ErrorMessage_a6ne ErrorMessage_a6ne
                                                 -> GHC.Types.Type))
  = forall arg_a6no. SameKind (Apply (:$$:$) arg_a6no) ((:$$:$$) arg_a6no) =>
    (:$$:$###)

-- For ($$:)
[...]

PR #230, opened 2017-07-12
The great singletons colon cleanup of 2017

### LINKED ISSUE #163: Support promoting symbolic types
GHC 8 can, so we should too. But see [the comment in `singInfixDecl`](https://github.com/goldfirere/singletons/blob/0e6c6218c3b19f945fc4f464d639e4e26759e6a8/src/Data/Singletons/Single/Fixity.hs#L10-L17) before proceeding too far, as I don't know how to handle that function if we can promote symbolic types.

RyanGlScott:
This is an imposingly large patch, but the end result is a nice cleanup of how `singletons` deals with infix names. I'm quite happy with the result.

The big change is that when we promote/single infix things, we prepend colons at the front far less often than before. The `CHANGES` file details exactly what changed, but as a general rule, an infix function/datatype name's promoted counterpart now has the same name. This makes the rules for singling infix names easier to remember: 90% of the time, you just prepend `%` (the other 10% are infix constructors, which _must_ start with a colon, so their singletons begin with `:%`).

This avoids some nasty ambiguities involving defunctionalization symbols, kind inference constructors (#200), infix datatype names, and infix class names (#226). I've added tests for these to ensure that they work.

I made some surprising findings during this:

1. We need to add two special cases for particular infix functions: `(.)` and `(!)`. That's simply because you can't define something named `(.)` or `(!)` at the type level. Bummer. I've chosen `(:.)` and `(:!)` as their promoted equivalents (luckily, `(:.)` already exists in `singletons`, so that's not so hard to remember).
2. We now promote `(&&)` and `(||)` to exactly those same names, instead of `(:&&)` and `(:||)`. Since their promoted counterparts now clash with type families from `Data.Type.Bool`, I decided to just re-export the ones from `Data.Type.Bool` in `singletons` (and come up with defunctionalization symbols/singleton functions for them as well).
3. I needed some extra functionality from `th-desugar` to properly disambiguate type-level and value-level infix things in quasiquotes. See `Note [Using dsReifyTypeNameInfo when promoting instances]` for a detailed explanation.

Fixes #163, #200, and #226.


RyanGlScott:
I just realized Travis won't be able to build this, since I used my own fork of `th-desugar` here to push through the submodule changes...

src/Data/Singletons/Promote.hs
```diff
@@ -391,15 +436,25 @@ promoteMethod m_subst sigs_map (meth_name, meth_rhs) = do
...
-        mb_info <- dsReify proName
-        case mb_info of
-          Just (DTyConI (DOpenTypeFamilyD (DTypeFamilyHead _ tvbs mb_res_ki _)) _)
-            -> let arg_kis = map (default_to_star . extractTvbKind) tvbs
-                   res_ki  = default_to_star (resultSigToMaybeKind mb_res_ki)
-               in return (arg_kis, res_ki)
-          _ -> fail $ "Cannot find type annotation for " ++ show proName
+        let mk_meth_ty = extract_meth_ty (dsReify proName)
+                     <|> extract_meth_ty (dsReifyTypeNameInfo proName)
+                     -- See Note [Using dsReifyTypeNameInfo when promoting instances]
```
RyanGlScott: After thinking about this some more, I think there might be a cleaner way to do this. I just noticed that `th-desugar` defines a `mkTypeName` function, which attempts to create a `Name` that's specifically in the type namespace. I believe this is exactly what I need, since I'm attempting to look up a _promoted_ method name, which should certainly live in the type namespace. If I modify `promoteValNameLhs` to use `mkTypeName`, then I think it should just work.

(The `th-desugar` patch to add `lookup{Value,Type}NameWithLocals` might be useful own its own merits, however.)

RyanGlScott: Indeed, it's much simpler now. I still need to bump the `th-desugar` submodule, though, since I'm using [`mkTypeNameWithLocals`](https://github.com/goldfirere/th-desugar/pull/62#issuecomment-314827010).

RyanGlScott: Hold the phone. We definitely don't want to go with the `mkTypeNameWithLocals` approach - at least not while [GHC Trac #13968](https://ghc.haskell.org/trac/ghc/ticket/13968) is around. That's because TH will try to splice in exact `Name`s for promoted things (instead of `Name`s that were created from just `mkName`). When combined with GHCi, this can lead to horrifying results like this that I just observed:

```
λ> ; $(singletons [d| f :: Bool; f = True |])
<interactive>:4:5-41: Splicing declarations
    singletons
      [d| f_a206O :: Bool
          f_a206O = True |]
  ======>
    f_a207e :: Bool
    f_a207e = True
    type FSym0 = interactive:Ghci1.F
    type family interactive:Ghci1.F :: Bool where
      = TrueSym0
    sF :: Sing (FSym0 :: Bool)
    sF = STrue
λ> ; $(singletons [d| f :: Bool; f = True |])
<interactive>:5:5-41: Splicing declarations
    singletons
      [d| f_a207Z :: Bool
          f_a207Z = True |]
  ======>
    f_a208p :: Bool
    f_a208p = True
    type FSym0 = F
    type family F :: Bool where
      = TrueSym0
    sF :: Sing (FSym0 :: Bool)
    sF = STrue

<interactive>:5:5: error: Illegal binding of built-in syntax: F
```

I'll revert the `mkTypeNameWithLocals` changes here, since the previous iteration of this pull request didn't suffer from this problem.

RyanGlScott:
Gah, I just realized that there's another thing you can define at the value level, but not at the type level: `(!)`:

```
λ> let (!) = 4
λ> type (!) = Int

<interactive>:36:8: error: parse error on input ‘)’
```

I should probably make a special case for that as well.

RyanGlScott:
@goldfirere, how does this look? I'm hesitant to merge this without at least one other pair of eyeballs looking over it, since it touches so many parts of the library.

src/Data/Singletons/Promote.hs
```diff
@@ -325,14 +328,56 @@ promoteInstanceDec meth_sigs
...
+    type (==) (x :: a) (y :: a) :: Bool
+
+Notice how both of these classes have an identifier named (==), one at the
+value level, and one at the type level. Now imagine what happens when you
+attempt to promote this Template Haskell declaration:
+
+   [d| f :: Bool
+       f = () == () |]
+
+Because it's quasiquoted, Template Haskell will attempt to reify a raw "=="
```
goldfirere: That's not a *quasi*quote. It's just a TH quote.

goldfirere: What's a "raw" name? When you quote that syntax, you'll correctly get a term-level `==`. Because names from quotes have namespace info attached to them (I'm pretty sure), I don't think this call to `reify` is ambiguous.

I'm also dubious of referring to GHC#13961, when that ticket is closed "invalid".

And, what does this example have to do with the code above? That's all about working with instance declarations, and your example doesn't have an instance declaration. So I'm very confused.

src/Data/Singletons/Promote.hs
```diff
@@ -325,14 +328,56 @@ promoteInstanceDec meth_sigs
...
+Name. But due to GHC Trac #13961, Template Haskell has to arbitrarily choose
+between reifying the info for the value-level (==) and the type-level (==), and
+in this case, it happens to pick the value-level (==) info. We want the type-level
+(==) info, however, because we care about the promoted version of (==).
+
+Fortunately, there's a serviceable workaround. After calling dsReify, if we happen
```
goldfirere: I'm dubious of this whole thing. Shouldn't we always want type-level info? (Or maybe it's term-level info... but it should always be one thing.)

Put another way: you have 4 alternatives above. For me to be convinced of correctness here, I would want examples of why each of the 4 is necessary.

src/Data/Singletons/Promote.hs
goldfirere: I'm sorry. I've misunderstood your point here. Always using `mkTypeName` seems very reasonable to me. What's so horrible about the output above?

src/Data/Singletons/Single/Fixity.hs
```diff
@@ -7,21 +7,42 @@ import Data.Singletons.Util
...
-    -- is it a tycon name or a datacon name??
-    -- it *must* be a datacon name, because symbolic tycons
-    -- can't be promoted. This is terrible.
-    DInfixD fixity (singDataConName name)
-  | otherwise = DInfixD fixity (singValName name)
+  | data_con_name
+  , ':':_ <- nameBase name
+    -- We must be careful with infix constructor names here. It's not
+    -- sufficient to check if a name begins with a colon to conclude that it
+    -- is in fact the name of a constructor. For instance, :*: could name a
```
goldfirere: Will a type-level name ever come here? I would expect this function to be called only on term-level names, making this extra logic unnecessary. If a type-level name *does* make it here, perhaps that's a bug.

src/Data/Singletons/Promote.hs
RyanGlScott: By "4 alternatives", I assume you mean this:

```haskell
    extract_tvb_names (dsReify pClsName)
<|> extract_tvb_names (dsReify cls_name)
<|> extract_tvb_names (dsReifyTypeNameInfo pClsName)
<|> extract_tvb_names (dsReifyTypeNameInfo cls_name)
```

`dsReify pClsName` and `dsReify cls_name` were there before this PR. We have to fall back on `dsReifyTypeNameInfo pClsName` and `dsReifyTypeNameInfo cls_name` in the event that the promoted version of a Name also happens to have a value-level counterpart (see my earlier comment about promoting `==`).

src/Data/Singletons/Single/Fixity.hs
goldfirere: These squidgy fixity declarations are terrible. (As in, GHC's treatment of them.) Perhaps a cleaner design is to have th-desugar figure out whether a fixity declaration is for a type-level thing or a term-level thing and then singletons has an easier job. But until then, this'll do.

RyanGlScott: Indeed, I would love a nice `Name -> DsM NameSpace` function for this purpose. I think I'll open a `th-desugar` issue for this.

RyanGlScott: https://github.com/goldfirere/th-desugar/issues/63

src/Data/Singletons/Promote.hs
goldfirere: Would the bottom two always work? And, is it possible to write a `mkTypeName` function that's like `mkName` but gives it a type-level namespace? If not, perhaps TH should be improved to allow this -- it's a very sensible thing to want.

src/Data/Singletons/Single/Fixity.hs
goldfirere: The "pure" variant I mentioned in goldfirere/th-desugar#63 should work in your case here. I think.

src/Data/Singletons/Promote.hs
RyanGlScott: > Would the bottom two always work?

I'm not sure I understand the question, but the reason the latter two would produce different results from the former two is precisely because `dsReifyTypeNameInfo` uses `lookupTypeNameWithLocals` to guide the reification process, so in the event of a class name being conflated with a value name, `dsReifyTypeNameInfo` will pick the class name.

RyanGlScott: > And, is it possible to write a `mkTypeName` function that's like `mkName` but gives it a type-level namespace?

Currently, I don't think so. The only constructor of [`NameFlavour`](http://hackage.haskell.org/package/template-haskell-2.12.0.0/docs/Language-Haskell-TH-Syntax.html#t:NameFlavour) that bundles a `NameSpace` is `NameG`, but that assumes the name is fully qualified.

> If not, perhaps TH should be improved to allow this -- it's a very sensible thing to want.

Perhaps. You can achieve much the same thing by using `lookupTypeName`, however, so I'm not keen on spending more time pursuing this featurette.

RyanGlScott: Actually, I believe I did misunderstand your "Would the bottom two always work?" question. I think I should have interpreted that as "Can you scrap the top two alternatives in favor of just the bottom two alternatives?" And the answer to that question is yes, happily enough. I've applied this change just now.

src/Data/Singletons/Single/Fixity.hs
RyanGlScott: It turns out `reifyNameSpace` needs to be monadic to work properly in all cases, but regardless, `reifyNameSpace` made the implementation of `singInfixDecl` _much_ cleaner. (See https://github.com/goldfirere/th-desugar/pull/64 for the `th-desugar` PR for `reifyNameSpace`.)

RyanGlScott:
GitHub weirdness aside, I believe I've addressed all of the outstanding concerns here. Should I change anything else?

PR #241, opened 2017-07-31
Fix #208 by adding a ShowSing class

### LINKED ISSUE #208: Generate Show instances for every Sing instance
#179 added custom `Show` instances for `SNat` and `SSym` (as well as `SomeSing Nat` and `SomeSing Symbol`). These are incredibly helpful for debugging purposes. I only wish there were more `Show` instances for other `Sing` instances!

Let's make this happen. I propose generating two additional declarations for each `Sing` instance ('ll use `[]` as an example):

1. Generate a `Show` instance for the `Sing` type. For instance:

   ```haskell
   instance (Show (Demote a), SingKind a, Typeable a) => Show (Sing (x :: [a])) where
     showsPrec p SNil
       = showParen (p > 10)
         ( showString "SNil"
           . showString " @"
           . showsPrec 11 (typeOf (undefined :: a))
         )
     showsPrec p (SCons x xs)
       = showParen (p > 10)
         ( showString "SCons"
           . showString " @"
           . showsPrec 11 (typeOf (undefined :: a))
           . showString " @"
           . showsPrec 11 (fromSing x)
           . showString " @"
           . showsPrec 11 (fromSing xs)
         )
   ```

   Note that coming up with the instance context would be very similar to that of `SingKind` instances, except we must also use additional `Show (Demote a)` and `Typeable a` constraints.

   Example usages (assuming we live in a post-#206 world):

   ```
   λ> SCons @Nat @0 @'[] sing sing
   SCons @Nat @0 @[]
   λ> SCons @[Nat] @'[0] @'[] sing sing
   SCons @[Nat] @[0] @[]
   ```
2. Derive the corresponding `Show` instance for `SomeSing`. For instance:

   ```haskell
   deriving instance (Show (Demote a), SingKind a, Typeable a) => Show (SomeSing [a])
   ```

RyanGlScott:
This implements the ideas laid out in #208. Some highlights:

* There is now a `Data.Singletons.ShowSing` module that exports the `ShowSing` class.
* The Template Haskell machinery has been updated so that calling `$(singletons [d| ... |])` on a datatype with a derived `Show` instance will also generate `ShowSing` and `Show` instances for the singleton type. (See `Note [DerivedDecl]`.)
* Along the way, I found it convenient to make deriving `Show` instances for empty datatypes work (instead of simply erroring as before), so I opted to do this.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -63,6 +63,9 @@ next
 
 * Add promoted and singled versions of `Show`, including `deriving` support.
 
+* Add a `ShowSing` class, which facilitates the ability to write `Show` instances
+  for `Sing` instances.
+
 * Permit derived `Ord` instances for empty datatypes.
 
 * Permit standalone `deriving` declarations.
--- a/README.md
+++ b/README.md
@@ -213,12 +213,12 @@ addition, there is a `ShowSing` class provided in the
 `Show` instances for `Sing` instances.
 
 What is the difference between the two? Let's use the `False` constructor as an
-example. If you used the `PShow Bool`, then the output of calling `Show_` on
-`False` is `"False"`, much like the value-level `Show Bool` instance (similarly
-for the `SShow Bool` instance). However, the `ShowSing Bool` instance is
-intended for printing the value of the _singleton_ constructor `SFalse`, so
-calling `showsSingPrec 0 SFalse` yields `"SFalse"` (simiarly for the
-`Show (Sing (SBool z))` instance).
+example. If you used the `PShow Bool` instance, then the output of calling
+`Show_` on `False` is `"False"`, much like the value-level `Show Bool` instance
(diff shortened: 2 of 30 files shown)


RyanGlScott:
GitHub seems might confused for some reason. Let me see if closing and re-opening makes a difference.

RyanGlScott:
How does this look, @goldfirere?

goldfirere:
Not sure what you're asking here. If it looks good to you, I'm happy too.

RyanGlScott:
OK. I just wanted to make sure this passed a routine inspection, since this touches so many files in `singletons`.

goldfirere:
Ah. This is a large patch. A quick breeze-through showed up nothing alarming. I'd say to commit -- we can always correct later if there's a problem. Thanks.

PR #259, opened 2017-10-11
Promote the type synonyms in Data.Singletons

### LINKED ISSUE #257: Promote function that constructs functions
I want to express some kind of "map" (each element has it's own function) of an heterogenous List and I don't know how I can express the relevant type-level function:

```haskell
$(promoteOnly [d|
    toArrow :: [Type] -> [Type] -> [Type -> Type]
    toArrow inputs outpus = zipWith (->) inputs outpus
  ])
```

I hope my question is sensible and I didn't mix up the terms!

RyanGlScott:
In https://github.com/goldfirere/singletons/issues/257#issuecomment-335919128, I lamented the lack of defunctionalization symbols for `(~>)`, so this PR adds them, as well as symbols for the other type synonyms in `Data.Singletons`.

Along the way, I realized that we were duplicating the export list for `Data.Singletons` in `Data.Singletons.Internal`, which seems silly, so I axed the explicit export list in `Data.Singletons.Internal`. (No one will notice, after all, since it's internal.)

Fixes #257.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -111,6 +111,9 @@ next
 
 * Add a `demote` function, which is a convenient shorthand for `fromSing sing`.
 
+* Export defunctionalization symbols for `SameKind, `KindOf`, `(~>)`, and `(@@)`
+  from `Data.Singletons`.
+
 2.3
 ---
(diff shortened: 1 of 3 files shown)


goldfirere:
LGTM. Thanks!

Issue #267, opened 2017-10-20
Can't promote a function returning String with a type signature


boothead:
If I have the following:

```
   $(singletons [d|
       data EventScans
        = PassThrough
        | SuiteProgress

       -- cn :: EventScans -> String
       cn PassThrough = "all-events"
       cn SuiteProgress = "suite-progress"
    |])
```
I can't un-comment the type signature for cn.

If I try and use the specific type signature above I get:

```
* Expected kind `[Char]', but `"all-events"' has kind `Symbol'
* In the type `"all-events"'
  In the type family declaration for `Cn' (haskell-stack-ghc)
```
If I leave the signature off I get `cn :: IsString t => EventScans  -> t` as the inferred type.

If I try to use the inferred type in the TH block I get:

```
Variable `a_agPb' used as both a kind and a type
 Did you intend to use TypeInType? (haskell-stack-ghc)
Not in scope: type constructor or class `SIsString'
  Perhaps you meant `IsString' (imported from Data.String) (haskell-stack-ghc)
```
I've tried this with lts-9.2 (and singletons-2.1) and I've also updated to singletons-2.2

There's a stack overflow question here: https://stackoverflow.com/questions/46834285/how-do-i-promote-a-function-that-returns-a-string-with-singletons I'm opening this issue in case this is a bug, it looks a bit similar to #247

RyanGlScott:
OK... and what happens is you enable `TypeInType`? :)

boothead:
I get an error on this class in a different part of the code:

```
class HasProcessor (api :: *) processor where
  type ProcessorT api processor (m :: * -> *) :: *
  server :: Proxy api -> Proxy processor -> Processor api processor

--  Operator applied to too few arguments: *
```

RyanGlScott:
Well, yes. That's to be expected—with `TypeInType` enabled, you have to explicitly import `*` from `Data.Kind`.

Let me put this another way: can you post a complete code example and highlight what you think the issue with `singletons` is? I can't figure out what you think the bug is.

boothead:
The concrete thing I'm asking about is why I can't specify the type signature of the cn function above. If I leave the type signature of the function the code compiles. If I enable `TypeInType` and add the type signature I get the following error:

```
Not in scope: type constructor or class `SIsString'
```

So my question is, is this a bug or user error with promoting a function? Should I expect to be able to include a type signature here, or just leave the type signature off?

RyanGlScott:
[...]

Issue #248, opened 2017-09-04
Errors in promoted functions


SamuraiJack:
If I'm correct "singletons" uses its own "Error" type for reporting error, which does not result in type error during compilation, but just silently wraps the type into the "Error".

```haskell
$(singletons [d|
    a :: Bool

    a = head []

    |])

data Some s = Some

b :: Some A

b = Some
```
compiles fine, its just the type is weird:
```haskell
> :t b
b :: Some (Error "Data.Singletons.List.head: empty list")
```
Should the `TypeError` facility from https://hackage.haskell.org/package/base-4.10.0.0/docs/GHC-TypeLits.html be used instead? It results in actual compilation error, with user message. Or am I missing something?

goldfirere:
Sounds good to me. @RyanGlScott ?

RyanGlScott:
This ticket largely overlaps with #160. To summarize that issue:

1. There's a design choice to be had here. Picking `Error` gives you a type family that doesn't reduce, whereas `TypeError` produces a compile-time exception. There are situations where you might want one or the other, so I'm hesitant to make all uses of `error` promote to `TypeError` unilaterally.
2. Given that, a natural inclination would be to create a term-level function `typeError :: ErrorMessage -> a` which promotes to the `TypeError` type family. This _almost_ works in today's `singletons`, but there's a really annoying obstacle in the form of the [`Text` constructor for `ErrorMessage`](https://www.stackage.org/haddock/lts-9.3/base-4.9.1.0/GHC-TypeLits.html#t:ErrorMessage):

   ```haskell
     | Text Symbol
   ```

   As chronicled in #82 (and #247, more recently), we can't use the `Text` symbol at the value level, only the type level. We could introduce a separate `ErrorMessageTerm` data type with a constructor `TextTerm :: Text -> ErrorMessageTerm`, but even if we did this, we still wouldn't be able to write the same code at the value and type level, which stinks.

   If we could redesign `TypeError` from scratch, we could make `Text` parameterized over a symbol type, as is done in https://github.com/goldfirere/singletons/issues/247#issuecomment-326762302. But I doubt we'd get away with that today.

goldfirere:
I had a niggling feeling things were more complicated than I thought -- which is why I wanted your opinion. Thanks for the explanation.

SamuraiJack:
"typeError" sounds good for me, with some form of workaround for a `Text/Symbol` issue..

SamuraiJack:
[...]

PR #268, opened 2017-10-20
Support OverloadedStrings in singletons

### LINKED ISSUE #247: Promote IsString
Hello,

I'm trying to singletonize the data structure with String/Text, using singletons 2.2. I'm planning to re-use this structure (along with associated methods) on both term and type level.

This fails with: `Could not deduce: DemoteRep Text ~ Text`
```haskell
$(singletons [d|
    data Some = Some {
        someName       :: Text
    } 
    |])
```

I've tried a workaround from #76. It does not compile with singletons 2.2 already, so I changed it a bit (replaced proxies)

This still fails however, with `Could not deduce: DemoteRep FooTerm ~ FooTerm`

```haskell
data Foo = Bar Symbol
data FooTerm = BarTerm Text   -- Integer is the closest term-level equiv.

data instance Sing (x :: Foo) where
  SBar :: Sing n -> Sing (Bar n)

instance SingKind Foo where
  type DemoteRep Foo = FooTerm
  fromSing (SBar n) = BarTerm $ pack (fromSing n)
  toSing (BarTerm n) = case toSing (unpack n) of
               SomeSing n' -> SomeSing (SBar n')

$(singletons [d|

    data Some = Some {
        someName       :: FooTerm
    } 
    |])
```

My idea is to define a custom singleton for "FooTerm" and then re-use it in the code being promoted (as I'm planning to add functions, working with "Some" and "FooTerm"). I'll only need the string equality comparison from the operations over the FooTerm.

Any advise?

RyanGlScott:
This has been requested enough times (see #247, #248, and #267) that I decided to just implement it.

Now, when the `-XOverloadedStrings` extension is enabled, occurrences of string literals will be promoted/singled using `fromSing` (e.g., you'd promote `"foo"` to `FromSing "foo"` and single `"foo"` to `sFromSing (sing :: Sing "foo")`).

Fixes #247.

### DIFF
--- a/src/Data/Singletons/Prelude/IsString.hs
+++ b/src/Data/Singletons/Prelude/IsString.hs
@@ -1,0 +1,43 @@
+{-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeInType #-}
+
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Data.Singletons.Prelude.IsString
+-- Copyright   :  (C) 2017 Ryan Scott
+-- License     :  BSD-style (see LICENSE)
+-- Maintainer  :  Richard Eisenberg (rae@cs.brynmawr.edu)
+-- Stability   :  experimental
+-- Portability :  non-portable
+--
+-- Defines and exports a promoted and singled version of the 'IsString'
+-- type class from "Data.String".
+----------------------------------------------------------------------------
+
+module Data.Singletons.Prelude.IsString (
+  PIsString(..), SIsString(..),
+
+  -- ** Defunctionalization symbols
+  FromStringSym0, FromStringSym1
+  ) where
+
+import Data.Singletons.Single
+import Data.Singletons.TypeLits ()   -- for the IsString instance!
+import GHC.TypeLits (Symbol)
+
+$(singletonsOnly [d|
+  -- -| Class for string-like datastructures; used by the overloaded string
+  --    extension (-XOverloadedStrings in GHC).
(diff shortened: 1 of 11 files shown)


src/Data/Singletons/Prelude/IsString.hs
```diff
@@ -0,0 +1,43 @@
...
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeInType #-}
+
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Data.Singletons.Prelude.IsString
+-- Copyright   :  (C) 2017 Ryan Scott
+-- License     :  BSD-style (see LICENSE)
+-- Maintainer  :  Richard Eisenberg (rae@cs.brynmawr.edu)
```
goldfirere: You know, this line is really becoming a lie... :)

src/Data/Singletons/Single.hs
```diff
@@ -666,6 +667,13 @@ singLit (IntegerL n)
...
                    return $ DVarE sNegateName `DAppE` sLit
+singLit (StringL str) = do
+  let sing_str_lit = DVarE singMethName `DSigE`
+                     (singFamily `DAppT` DLitT (StrTyLit str))
+  os_enabled <- qIsExtEnabled LangExt.OverloadedStrings
+  pure $ if os_enabled
+         then DVarE sFromStringName `DAppE` sing_str_lit
+         else sing_str_lit
 singLit lit = do
   prom_lit <- promoteLitExp lit
```
goldfirere: This will never work, I think. Remove.

RyanGlScott: OK, I've turned this line into an error.

Issue #273, opened 2017-11-10
Add SingI (& other) pattern synonyms


Icelandjack:
This would be useful

```haskell
pattern SingI :: () => SingI a => Π(a::k)
pattern SingI <- (singInstance -> SingInstance)
  where SingI = sing
```

----

Similarly I think the same should be added to [`Type.Reflection`](https://downloads.haskell.org/~ghc/master/libraries/html/base/Type-Reflection.html), where is best to propose that again?

```haskell
asTypeable :: TypeRep a -> Dict (Typeable a)
asTypeable rep =
  withTypeable rep
    Dict

pattern TypeRep :: () => Typeable a => TypeRep a
pattern TypeRep <- (asTypeable -> Dict)
  where TypeRep = typeRep
```

RyanGlScott:
I must be missing something obvious here, since I don't see what utility this provides, given that every `Sing` data family instance comes patterns (in the form of constructors) that accomplish this very thing. Can you give an example that demonstrates where this would be a win?

Icelandjack:
Anywhere where `withSingI` is useful (i.e. where you want to constraintify an explicit witness), using the continuation passing `with*` functions work fine but explicit pattern matching can lead to nicer code. 

These can be used as building blocks for other pattern synonyms.

```haskell
-- repl :: forall a n. SingI n => a -> Vec n a
-- repl a = case sing :: Π(n) of
--   SO    -> VO
--   SS n' -> a :> withSingI n' (repl a)

repl :: forall a n. SingI n => a -> Vec n a
repl a = case sing :: Π(n) of
  SO       -> VO
  SS SingI -> a :> repl a
```

Similarly any of these may be useful

### `AsSing`

```haskell
asSing :: SingKind k => Demote k -> SomeSing k
asSing demote =
  withSomeSing demote
    SomeSing

pattern AsSing :: SingKind k => forall (a :: k). Π(a::k) -> Demote k
pattern AsSing sing <- (asSing -> SomeSing sing)
  where AsSing sing = fromSing sing
```

uses

```haskell
-- toSing :: N -> SomeSing N
-- toSing O     = SomeSing SO
-- toSing (S n) = withSomeSing n (\n' -> SomeSing $ SS n')
toSing_ :: N -> SomeSing N
toSing_ = \case
  O -> SomeSing SO
  S (AsSing sn) ->
    SomeSing (SS sn)
```

### `Demoted`

```haskell
[...]

PR #276, opened 2017-11-10
Added SingI pattern synonym.

Icelandjack:
Solves the original question of #273.

### DIFF
--- a/src/Data/Singletons.hs
+++ b/src/Data/Singletons.hs
@@ -6,10 +6,7 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE TypeInType #-}
-{-# LANGUAGE CPP #-}
-#if defined(__GLASGOW_HASKELL__) && __GLASGOW_HASKELL__ >= 710
 {-# LANGUAGE PatternSynonyms #-}
-#endif 
 {-# OPTIONS_GHC -Wno-orphans #-}
 
 -----------------------------------------------------------------------------
@@ -40,15 +37,12 @@ module Data.Singletons (
   SingI(..), SingKind(..),
 
   -- * Working with singletons
-  KindOf, SameKind,
-  SingInstance(..), SomeSing(..),
-  singInstance, withSingI, withSomeSing, singByProxy, demote,
-#if defined(__GLASGOW_HASKELL__) && __GLASGOW_HASKELL__ >= 710
+  KindOf, SomeSing(..), SameKind,
+  SingInstance(..), singInstance, 
   pattern SingI,
-#endif
 
-  singByProxy#,
-  withSing, singThat,
+  withSingI, withSomeSing, singByProxy, demote,
+  singByProxy#, withSing, singThat,
 
   -- ** Defunctionalization
   TyFun, type (~>),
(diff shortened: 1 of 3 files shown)


src/Data/Singletons.hs
```diff
@@ -6,6 +6,10 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE TypeInType #-}
+{-# LANGUAGE CPP #-}
+#if defined(__GLASGOW_HASKELL__) && __GLASGOW_HASKELL__ >= 710
```
RyanGlScott: `singletons` only supports GHC 8.2, so there's no need for this CPP.

src/Data/Singletons.hs
```diff
@@ -39,6 +43,9 @@ module Data.Singletons (
   KindOf, SameKind,
   SingInstance(..), SomeSing(..),
   singInstance, withSingI, withSomeSing, singByProxy, demote,
+#if defined(__GLASGOW_HASKELL__) && __GLASGOW_HASKELL__ >= 710
```
RyanGlScott: Remove the CPP here.

src/Data/Singletons.hs
```diff
@@ -39,6 +43,9 @@ module Data.Singletons (
...
+  pattern SingI,
```
RyanGlScott: Given that (logically) this pattern synonym is an extension of `singInstance`, I would put this next to `singInstance` in the export list.

src/Data/Singletons/Internal.hs
```diff
@@ -2,7 +2,13 @@
              FlexibleContexts, FlexibleInstances,
              TypeFamilies, TypeOperators, TypeFamilyDependencies,
              UndecidableInstances, TypeInType, ConstraintKinds,
-             ScopedTypeVariables, TypeApplications, AllowAmbiguousTypes #-}
+             ScopedTypeVariables, TypeApplications, AllowAmbiguousTypes,
+             CPP #-}
+
+#if defined(__GLASGOW_HASKELL__) && __GLASGOW_HASKELL__ >= 710
```
RyanGlScott: Remove the CPP here.

src/Data/Singletons/Internal.hs
```diff
@@ -49,12 +55,24 @@ data family Sing (a :: k)
...
 -- | A 'SingI' constraint is essentially an implicitly-passed singleton.
 -- If you need to satisfy this constraint with an explicit singleton, please
--- see 'withSingI'.
+-- see 'withSingI' or pattern match using the 'SingI' pattern synonym.
 class SingI (a :: k) where
   -- | Produce the singleton explicitly. You will likely need the @ScopedTypeVariables@
   -- extension to use this method the way you want.
   sing :: Sing a
 
+#if __GLASGOW_HASKELL__ >= 710
```
RyanGlScott: Remove the CPP here.

src/Data/Singletons/Internal.hs
```diff
@@ -49,12 +55,24 @@ data family Sing (a :: k)
...
+-- | Pattern matching on an explicit @Sing a@ to get an implicit
+-- @SingI a@ constraint.
+--
+-- /Since: 2.4/
+#if __GLASGOW_HASKELL__ >= 800
```
RyanGlScott: Remove the CPP here.

RyanGlScott:
Thanks!

PR #277, opened 2017-11-11
Added AsSing and documentation.

Icelandjack:
From #273

```haskell
pattern AsSing :: SingKind k => forall (a :: k). Sing a -> Demote k
pattern AsSing sing <- ((\demote -> withSomeSing demote SomeSing) -> SomeSing sing)
  where AsSing sing = fromSing sing 
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -122,15 +122,16 @@ next
 * Export defunctionalization symbols for `Demote`, `SameKind, `KindOf`, `(~>)`,
   `Apply`, and `(@@)` from `Data.Singletons`.
 
-* Add an explicitly bidirectional pattern synonym `SingI`. Pattern
+* Add explicitly bidirectional pattern synonym `SingI`. Pattern
   matching on `SingI` brings a `SingI ty` constraint into scope from a
   singleton `Sing ty`.
 
-* Add a pattern synonym `AsSing`. Pattern matching on any demoted
-  (base) type gives us the corresponding singleton.
+* Add explicitly bidirectional pattern synonym `AsSing`. Pattern
+  matching on any demoted (base) type gives us the corresponding
+  singleton.
 
-* Add an explicitly bidirectional pattern synonym `SLambda2`,
-  `SLambda{3..8}`. Pattern matching on any defunctionalized singleton
+* Add explicitly bidirectional pattern synonym
+  `SLambda{2..8}`. Pattern matching on any defunctionalized singleton
   yields a term-level Haskell function on singletons. Equals the
   expression `singFun2`, `singFun{3..8}` and accessor `unSingFun2`,
   `unSingFun{3..8}`.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -122,10 +122,18 @@ next
 * Export defunctionalization symbols for `Demote`, `SameKind, `KindOf`, `(~>)`,
   `Apply`, and `(@@)` from `Data.Singletons`.
 
-* Add a pattern synonym `SingI`. Pattern matching on `SingI` brings a
- `SingI ty` constraint into scope from a singleton `Sing ty`.
+* Add an explicitly bidirectional pattern synonym `SingI`. Pattern
+  matching on `SingI` brings a `SingI ty` constraint into scope from a
+  singleton `Sing ty`.
 
-* Add a pattern synonym `AsSing`. Pattern matching on any demoted (base) type gives us the corresponding singleton.
+* Add a pattern synonym `AsSing`. Pattern matching on any demoted
+  (base) type gives us the corresponding singleton.
+
+* Add an explicitly bidirectional pattern synonym `SLambda2`,
+  `SLambda{3..8}`. Pattern matching on any defunctionalized singleton
+  yields a term-level Haskell function on singletons. Equals the
+  expression `singFun2`, `singFun{3..8}` and accessor `unSingFun2`,
+  `unSingFun{3..8}`.
 
 2.3
 ---
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -125,6 +125,8 @@ next
 * Add a pattern synonym `SingI`. Pattern matching on `SingI` brings a
  `SingI ty` constraint into scope from a singleton `Sing ty`.
 
+* Add a pattern synonym `AsSing`. Pattern matching on any demoted (base) type gives us the corresponding singleton.
+
 2.3
 ---
 * Documentation clarifiation in `Data.Singletons.TypeLits`, thanks to @ivan-m.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -122,19 +122,17 @@ next
 * Export defunctionalization symbols for `Demote`, `SameKind, `KindOf`, `(~>)`,
   `Apply`, and `(@@)` from `Data.Singletons`.
 
-* Add explicitly bidirectional pattern synonym `SingI`. Pattern
-  matching on `SingI` brings a `SingI ty` constraint into scope from a
+* Add explicitly bidirectional pattern synonym `Sing`. Pattern
+  matching on `Sing` brings a `SingI ty` constraint into scope from a
   singleton `Sing ty`.
 
-* Add explicitly bidirectional pattern synonym `AsSing`. Pattern
+* Add explicitly bidirectional pattern synonym `FromSing`. Pattern
   matching on any demoted (base) type gives us the corresponding
   singleton.
 
 * Add explicitly bidirectional pattern synonym
   `SLambda{2..8}`. Pattern matching on any defunctionalized singleton
-  yields a term-level Haskell function on singletons. Equals the
-  expression `singFun2`, `singFun{3..8}` and accessor `unSingFun2`,
-  `unSingFun{3..8}`.
+  yields a term-level Haskell function on singletons.
 
 2.3
 ---
--- a/src/Data/Singletons.hs
+++ b/src/Data/Singletons.hs
@@ -35,6 +35,7 @@ module Data.Singletons (
   Sing(SLambda, applySing), (@@),
 
   SingI(..), SingKind(..),
+  pattern AsSing,
 
   -- * Working with singletons
   KindOf, SameKind,
--- a/src/Data/Singletons.hs
+++ b/src/Data/Singletons.hs
@@ -39,8 +39,8 @@ module Data.Singletons (
   -- * Working with singletons
   KindOf, SameKind,
   SingInstance(..), SomeSing(..),
-  singInstance, pattern SingI, withSingI,
-  withSomeSing, pattern AsSing,
+  singInstance, pattern Sing, withSingI,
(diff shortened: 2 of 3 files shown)


Icelandjack:
The example isn't great

```haskell
singAnd :: Bool -> Bool -> SomeSing Bool
singAnd bool1 bool2
  | AsSing singBool1 <- bool1
  , AsSing singBool2 <- bool2
  = SomeSing (singBool1 %:&& singBool2)
```

instead of

```haskell
singAnd bool1 bool2 =
  withSomeSing bool1 $ \sBool1 ->
    withSomeSing bool2 $ \sBool2 ->
      SomeSing (sBool1 %:&& sBool2)
```

in case anyone has a better example.

src/Data/Singletons/Internal.hs
```diff
@@ -97,6 +111,40 @@ class SingKind k where
 data SomeSing k where
   SomeSing :: Sing (a :: k) -> SomeSing k
 
+-- | A pattern synonym for going between a term and the corresponding
```
RyanGlScott: The corresponding what?

src/Data/Singletons/Internal.hs
```diff
@@ -97,6 +111,40 @@ class SingKind k where
...
+--
+-- As a constructor: this takes a singleton to its demoted (base)
+-- type.
+--
+-- >>> :t AsSing @Bool
```
RyanGlScott: IIRC, I believe you need to escape the `@` character (as `\@`) in order for Haddock to render it correctly.

src/Data/Singletons/Internal.hs
```diff
@@ -97,6 +111,40 @@ class SingKind k where
...
+-- False
+--
+-- As a pattern: It extracts a singleton from its demoted (base) type.
+--
+-- @
+-- singAnd :: Bool -> Bool -> SomeSing Bool
+-- singAnd bool1 bool2
+--   | AsSing singBool1 <- bool1
+--   , AsSing singBool2 <- bool2
+--   = SomeSing (singBool1 %:&& singBool2)
```
RyanGlScott: Why not write this directly like so?

```haskell
singAnd (AsSing singBool1) (AsSing singBool2) = SomeSing (singBool1 %&& singBool2)
```

Also, note that I wrote `%&&`, not `%:&&`: we don't use colons for singled infix operators anymore (after #230).

src/Data/Singletons/Internal.hs
```diff
@@ -56,8 +56,15 @@ class SingI (a :: k) where
...
--- implicit @SingI a@ constraint into scope.
+-- | A pattern synonym for implicit singletons.
+--
+-- As a pattern: We can match on an explicit @Sing a@ witness to bring
+-- an implicit @SingI a@ constraint into scope.
+--
+-- As a constructor: It constructs a singleton @Sing a@ as long as we
+-- have an implicit singleton constraint @SingI a@.
+--
+-- /Since: 2.4/
```
RyanGlScott: There's not much point in putting `Since` annotations in this library, since we completely break backwards compatibility every release.

src/Data/Singletons.hs
```diff
@@ -35,6 +35,7 @@ module Data.Singletons (
   Sing(SLambda, applySing), (@@),
 
   SingI(..), SingKind(..),
+  pattern AsSing,
```
RyanGlScott: I would this next to `withSomeSing` in the export list, similarly to now `pattern SingI` is currently next to `singInstance`.

src/Data/Singletons/Internal.hs
Icelandjack: I found better uses of `withSomeSing` but it wasn't easy to minimize them into an example

RyanGlScott: To be clear: I think the example you've given is fine. I'm just suggesting a more concise way to write it.

src/Data/Singletons/Internal.hs
Icelandjack: It is often useful to me to know when a feature was added, I'll remove it for consistency

src/Data/Singletons/Internal.hs
Icelandjack: Hm Kmett seems to use `\@` all over the place but it seems to work withou it here https://hackage.haskell.org/package/singletons-2.3.1/docs/Data-Singletons.html#v:singFun1

src/Data/Singletons/Internal.hs
Icelandjack: I misread. Updating to your version

CHANGES.md
```diff
@@ -122,8 +122,18 @@ next
...
-* Add a pattern synonym `SingI`. Pattern matching on `SingI` brings a
- `SingI ty` constraint into scope from a singleton `Sing ty`.
+* Add an explicitly bidirectional pattern synonym `SingI`. Pattern
+  matching on `SingI` brings a `SingI ty` constraint into scope from a
+  singleton `Sing ty`.
+
+* Add a pattern synonym `AsSing`. Pattern matching on any demoted
+  (base) type gives us the corresponding singleton.
+
+* Add an explicitly bidirectional pattern synonym `SLambda2`,
```
RyanGlScott: This sentence reads strangely. I would say: "Add explicitly bidirectional pattern synonyms `SLambda{2..8}`."

CHANGES.md
```diff
@@ -122,8 +122,18 @@ next
...
+  `SLambda{3..8}`. Pattern matching on any defunctionalized singleton
+  yields a term-level Haskell function on singletons. Equals the
```
RyanGlScott: I would remove this last sentence ("Equals the...") since it reads strangely and is more confusing than anything.

src/Data/Singletons.hs
```diff
@@ -53,6 +54,9 @@ module Data.Singletons (
   -- ** Defunctionalized singletons
   -- | When calling a higher-order singleton function, you need to use a
   -- @singFun...@ function to wrap it. See 'singFun1'.
+  pattern SLambda2, pattern SLambda3, pattern SLambda4, pattern SLambda5,
```
RyanGlScott: Please put these after the list of `singFun*` and `unSingFun*` functions, since the Haddocks above this line refer to `singFun*`, but now the entities directly below them _aren't_ `singFun*`, which could lead to confusion.

src/Data/Singletons/Internal.hs
```diff
@@ -56,8 +56,13 @@ class SingI (a :: k) where
   -- extension to use this method the way you want.
   sing :: Sing a
 
--- | Permits pattern matching on an explicit @Sing a@ value to bring an
--- implicit @SingI a@ constraint into scope.
+-- | An (explicitly bidirectional) pattern synonym for implicit singletons.
```
RyanGlScott: I'd opt to remove the parentheses around "(explicitly bidirectional)".

src/Data/Singletons/Internal.hs
```diff
@@ -97,6 +109,38 @@ class SingKind k where
 data SomeSing k where
   SomeSing :: Sing (a :: k) -> SomeSing k
 
+-- | An (explicitly bidirectional) pattern synonym for going between a
```
RyanGlScott: Similarly, remove the parentheses around (explicitly bidirectional) here.

src/Data/Singletons/Internal.hs
```diff
@@ -97,6 +109,38 @@ class SingKind k where
...
+-- term and the corresponding demoted term.
```
RyanGlScott: I would phrase this as "... for going between a singleton and the corresponding demoted term."

src/Data/Singletons/Internal.hs
```diff
@@ -178,6 +222,72 @@ infixl 9 @@
 newtype instance Sing (f :: k1 ~> k2) =
   SLambda { applySing :: forall t. Sing t -> Sing (f @@ t) }
 
+-- | An (explicitly bidirectional) pattern synonym for
```
RyanGlScott: Similarly, remove the parentheses around (explicitly bidirectional) here.

RyanGlScott: Also, these Haddocks are only attached to `SLambda2`, despite the fact that it equally applies to every `SLambda*` pattern synonym. In light of this, I would suggest turning this into a Haddock block, like so:

```haskell
-- (In the export list)
-- $SLambdaPatternSynonyms
  SLambda2, SLambda3, ...

{- $SLambdaPatternSynonyms

@SLambda{2...8}@ are explicitly bidirectional pattern synonyms for...
-}
```

src/Data/Singletons/Internal.hs
```diff
@@ -178,6 +222,72 @@ infixl 9 @@
...
+-- Haskell function on singletons @Sing (x :: k) -> Sing (y :: k') ->
+-- Sing (f \@\@ x \@\@ y)@ from a defunctionalised @Sing
+-- f@. Alternatively as a record field accessor
+--
+-- @
+-- applySing2 :: Sing (f :: k ~> k' ~> k'') -> SingFunction2 f
+-- @
+pattern SLambda2 :: forall f. SingFunction2 f -> Sing f
+pattern SLambda2 {applySing2} <- (unSingFun2 -> applySing2)
+  where SLambda2 lam2         = singFun2 lam2
```
RyanGlScott: Just for the sake of completeness, I'd add one `COMPLETE` pragma for each of these `SLambda*` pattern synonyms. (They won't ever kick in at the moment due to [Trac #14059](https://ghc.haskell.org/trac/ghc/ticket/14059), but it won't harm anything to add now in anticipation of that bug being fixed one day.)

src/Data/Singletons/Internal.hs
```diff
@@ -178,6 +222,72 @@ infixl 9 @@
...
+pattern SLambda6 {applySing6} <- (unSingFun6 -> applySing6)
+  where SLambda6 lam6         = singFun6 lam6
+
+pattern SLambda7 :: forall f. SingFunction7 f -> Sing f
+pattern SLambda7 {applySing7} <- (unSingFun7 -> applySing7)
+  where SLambda7 lam7         = singFun7 lam7
+
+pattern SLambda8 :: forall f. SingFunction8 f -> Sing f
+pattern SLambda8 {applySing8} <- (unSingFun8 -> applySing8)
+  where SLambda8 lam8         = singFun8 lam8
```
RyanGlScott: Again, I would put the code for these `SLambda*` pattern synonyms after the code for `singFun*` and `unSingFun*`.

Icelandjack:
I changed the names (from `SingI` to `Sing`) to match the name of the introduction form (`Sing = sing`), what mostly sold me on the idea was the equivalent pattern synonym for `TypeRep` / `Typeable` which I would now lean towards calling `TypeRep` (outside of the scope of this project).

Any thoughts?

RyanGlScott:
LGTM. The particular choice of name doesn't concern me too much—I think `Sing` is fine.

PR #278, opened 2017-11-15
Fix #271 by inferring instance contexts more intelligently

### LINKED ISSUE #271: singletons uses two completely different (and incompatible) type inference algorithms for deriving Eq and Ord
This doesn't compile:

```haskell
{-# LANGUAGE GADTs #-}
{-# LANGUAGE InstanceSigs #-}
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE TypeInType #-}
module Bug where

import Data.Kind
import Data.Singletons.TH

$(singletons [d| newtype Constant (a :: Type) (b :: Type) =
                    Constant a deriving (Eq, Ord) |])
```

```
$ /opt/ghc/8.2.1/bin/ghci Bug.hs
GHCi, version 8.2.1: http://www.haskell.org/ghc/  :? for help
Loaded GHCi configuration from /home/ryanglscott/.ghci

Bug.hs:15:3: error:
    • Could not deduce (SEq b)
        arising from the superclasses of an instance declaration
      from the context: SOrd a
        bound by the instance declaration at Bug.hs:(15,3)-(16,52)
      Possible fix:
        add (SEq b) to the context of the instance declaration
    • In the instance declaration for ‘SOrd (Constant a b)’
   |
15 | $(singletons [d| newtype Constant (a :: Type) (b :: Type) =
   |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
```

`-ddump-splices` reveals the reason why:

```haskell
    instance (SEq a_a7tf, SEq b_a7tg) =>
             SEq (Constant_a7td a_a7tf b_a7tg) where
      ...
    instance SOrd a_a7tf => SOrd (Constant_a7td a_a7tf b_a7tg) where
      ...
```

Notice that the `SEq` and `SOrd` instances have two completely different inferred instance contexts! For `SEq`, there's a redundant `SEq b` constraint, whereas the `SOrd` instance lacks this.

This leads one to wonder: why are there two completely different ways to infer the instance contexts for `SEq` and `SOrd`? And shouldn't we use the algorithm for `SOrd` for `SEq` as well, given that the current algorithm for `SEq` appears to be plain broken?

RyanGlScott:
`deriving Eq` wasn't using `inferConstraints`, which led to all sorts of shenanigans in #271. This is easily remedied.

Another problem documented in #271 was the fact that `inferConstraints` was fragile in the presence of GADTs, since each constructors' universally quantified type variables could be different from the data type constructor's type variables. To fix this, we apply a substitution on the inferred field types (see `Data.Singletons.Deriving.Infer`).

### DIFF
--- a/src/Data/Singletons/Deriving/Infer.hs
+++ b/src/Data/Singletons/Deriving/Infer.hs
@@ -15,14 +15,10 @@
 
 module Data.Singletons.Deriving.Infer ( inferConstraints, inferConstraintsDef ) where
 
-import Language.Haskell.TH (Name)
 import Language.Haskell.TH.Desugar
 import Data.Singletons.Util
 import Data.List
-import qualified Data.Map as Map
-import Data.Map (Map)
 import Data.Generics.Twins
-import qualified Data.Set as Set
 
 -- @inferConstraints cls inst_ty cons@ infers the instance context for a
 -- derived type class instance of @cls@ for @inst_ty@, using the constructors
@@ -84,8 +80,8 @@ inferConstraints pr inst_ty = fmap (nubBy geq) . concatMapM infer_ct
     -- for this, when gathering @field_tys@ (from step 1 in the above algorithm)
     -- we perform the following extra steps:
     --
-    -- 1(a). Take the return type of @con@ and unify it with @inst_ty@ (e.g.,
-    --       unify @Bar b@ with @Bar a@). Doing so will produce a substitution
+    -- 1(a). Take the return type of @con@ and match it with @inst_ty@ (e.g.,
+    --       match @Bar b@ with @Bar a@). Doing so will produce a substitution
     --       that maps the universally quantified type variables in the GADT
     --       (i.e., @b@) to the corresponding type variables in the data type
     --       constructor (i.e., @a@).
@@ -99,8 +95,16 @@ inferConstraints pr inst_ty = fmap (nubBy geq) . concatMapM infer_ct
       let field_tys = tysOfConFields fields
       field_tys' <- case mb_res_ty of
                       Nothing -> pure field_tys
-                      Just res_ty -> do subst <- unify res_ty inst_ty
-                                        pure $ map (substType subst) field_tys
+                      Just res_ty -> do
+                        res_ty'  <- expandType res_ty
+                        inst_ty' <- expandType inst_ty
+                        case matchTy YesIgnore res_ty' inst_ty of
+                          Nothing -> fail $ showString "Unable to match type "
+                                          . showsPrec 11 res_ty'
+                                          . showString " with "
+                                          . showsPrec 11 inst_ty'
+                                          $ ""
+                          Just subst -> traverse (substTy subst) field_tys
       pure $ map (pr `DAppPr`) field_tys'
 
 -- For @inferConstraintsDef mb_cxt@, if @mb_cxt@ is 'Just' a context, then it will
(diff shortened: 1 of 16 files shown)


tests/compile-and-dump/GradingClient/Database.ghc82.template
```diff
@@ -96,12 +96,12 @@ GradingClient/Database.hs:(0,0)-(0,0): Splicing declarations
                SNil)
       sCompare SZero (SSucc _) = SLT
       sCompare (SSucc _) SZero = SGT
-    instance SEq Nat where
+    instance SEq Nat => SEq Nat where
```
goldfirere: This looks very scary and wrong. Can it be avoided?

RyanGlScott:
How does this look now, @goldfirere?

RyanGlScott:
Ping?

RyanGlScott:
Ping?

src/Data/Singletons/Deriving/Infer.hs
```diff
@@ -13,16 +15,146 @@
...
+    --       quantified type variables of @con@ as necessary.
+    --
+    -- After this renaming, the algorithm will produce an instance context of
+    -- @Show a@ (since @b@ was renamed to @a@), as expected.
+    infer_ct :: DCon -> q DCxt
+    infer_ct (DCon _ _ _ fields mb_res_ty) = do
+      let field_tys = tysOfConFields fields
+      field_tys' <- case mb_res_ty of
+                      Nothing -> pure field_tys
+                      Just res_ty -> do subst <- unify res_ty inst_ty
```
goldfirere: You don't really want unification here, as you don't want to map variables that appear in the `inst_ty`. You want matching. And that's what the brand-new `Language.Haskell.TH.Desugar.Subst.matchTy` function does. Use that here (after expanding synonyms, as you've already done).
 
`matchTy` takes a first argument telling it whether or not to ignore kind signatures. Here, you can (`YesIgnore`), because GHC ensures that the result type is well-kinded. You don't have to worry about it.

src/Data/Singletons/Deriving/Infer.hs
```diff
@@ -13,16 +15,146 @@
...
+      pure $ map (pr `DAppPr`) field_tys'
+
+-- For @inferConstraintsDef mb_cxt@, if @mb_cxt@ is 'Just' a context, then it will
+-- simply return that context. Otherwise, if @mb_cxt@ is 'Nothing', then
+-- 'inferConstraintsDef' will infer an instance context (using 'inferConstraints').
+inferConstraintsDef :: DsMonad q => Maybe DCxt -> DPred -> DType -> [DCon] -> q DCxt
+inferConstraintsDef mb_ctxt pr inst_ty cons =
+  maybe (inferConstraints pr inst_ty cons) pure mb_ctxt
+
+-- A thin wrapper around @unify'@ which expand through type synonyms in the
```
goldfirere: All the code from here on down can be removed.

src/Data/Singletons/Util.hs
```diff
@@ -270,6 +270,8 @@ noExactTyVars = everywhere go
 substKind :: Map Name DKind -> DKind -> DKind
 substKind = substType
 
+-- | Non–capture-avoiding substitution. (If you want capture-avoiding
+-- substitution, use @substTy@ from "Language.Haskell.TH.Desugar.Expand".
```
goldfirere: Now from `...Subst` not `...Expand`

RyanGlScott:
Thanks for the review! I've applied your suggestions just now.

goldfirere:
Looks good. Thanks for the very detailed, helpful comments! Merge.

PR #279, opened 2017-11-21
Singkind

Icelandjack:
If you find this useful

```haskell
class    (SingKind k, Demote k ~ demoted) => SingKindOf demoted k
instance (SingKind k, Demote k ~ demoted) => SingKindOf demoted k 
```

It's a trick I would like to see more of, used for example by Kmett in [`hask`](https://github.com/ekmett/hask/blob/master/src/Hask/Category.hs#L93) that lets you specify associated type families 

```haskell
type Cat ob = ob -> ob -> Type

class (Category (Dom f), Category (Cod f)) => Functor (f :: i -> j) where
  type Dom f :: Cat i
  type Cod f :: Cat j
  fmap :: Dom f a b -> Cod f (f a) (f b)

class    (Functor f, Dom f ~ dom, Cod f ~ cod) => FunctorOf dom cod f
instance (Functor f, Dom f ~ dom, Cod f ~ cod) => FunctorOf dom cod f
```

letting you specify "normal" Haskell functors: `type PreludeFunctor = FunctorOf (->) (->)`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,9 @@ Changelog for singletons project
 
 next
 ----
+* Pattern synonym `SingI` for getting `SingI ty` constraint from a
+  singleton `Sing ty`.
+
 * `Demote Nat` is now `Natural` (from `Numeric.Natural`) instead of `Integer`.
   In accordance with this change, `Data.Singletons.TypeLits` now exposes
   `GHC.TypeNats.natVal` (which returns a `Natural`) instead of
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,9 +3,6 @@ Changelog for singletons project
 
 next
 ----
-* Pattern synonym `SingI` for getting `SingI ty` constraint from a
-  singleton `Sing ty`.
-
 * `Demote Nat` is now `Natural` (from `Numeric.Natural`) instead of `Integer`.
   In accordance with this change, `Data.Singletons.TypeLits` now exposes
   `GHC.TypeNats.natVal` (which returns a `Natural`) instead of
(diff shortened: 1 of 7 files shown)


vagarenko:
Why not just use constraint type synonym?

type SingKindOf demoted k = (SingKind k, Demote k ~ demoted)

int-index:
@vagarenko such a synonym cannot be partially applied.

RyanGlScott:
I'll admit that I find this whole thing to be of questionable utility. As far as I can tell, the only thing you "gain" from this is the ability to write `demoted` instead of `Demote k`. For... what purpose exactly?

-1 from me.

RyanGlScott:
And yes, if you want a shorthand for `(SingKind k, Demote k ~ demoted)`, I'd recommend following @vagarenko's advice and using `genDefunSymbols` on `type SingKindOf demoted k = (SingKind k, Demote k ~ demoted)`.

int-index:
>  ability to write demoted instead of Demote k

From what I see, the motivation is when `k` is an expression like `KindOf a` (rather than just a variable `k`), so it's more cumbersome to write.

That said, I don't have a use for this shorthand, personally, and I don't think that we should extend API surface of singletons further (beginner users are drowning in complexity without these additions).

RyanGlScott:
> From what I see, the motivation is when `k` is an expression like `KindOf a` (rather than just a variable `k`), so it's more cumbersome to write.

Ah, I see what you mean now. Yes, I think what @Icelandjack _really_ wants is the ability to write something like:

```
foo :: let demoted = Demote (KindOf a) in Sing a -> demoted
```

As in dependently typed languages. I find the class-based approach presented here to be a poor imitation of this feature though, and I concur with @int-index that I wouldn't consider this to pass the bar needed for inclusion in `singletons`.

Icelandjack:
@vagarenko Partial application as @int-index says + not requiring enabling an extension at call site.

It bundles type class / family together (which is useful in complicated constraints) but it can also avoid mentioning the `Demote` type family at all. 

Given that `Demote` is an injective type family, it can make GHC to infer the kind from the demoted type

```
>>> sing :: (SingKindOf Integer k, Typeable k) => Sing k
... :: Sing Nat
```

That being said I'm fine with closing this ticket

PR #281, opened 2017-11-23
Generalize the TestCoercion instance for Sing

RyanGlScott:
Currently, the `TestCoercion Sing` instance only works if `Sing` is of kind `Type -> Type`. But this can quite easily work over `Sing`s of any kind... why not do so?


goldfirere:
Indeed, why not. Merge when ready. Thanks!

Issue #82, opened 2014-05-31
Promote datatypes with arrows, among others


goldfirere:
In thinking about #78, I realized that fixing that bug opens up a whole new possibility for promotion.

Say a user wants to promote

```
data NatPred = MkNP (Nat -> Bool)
```

Currently, that just won't promote, because functions are different in types than they are in terms. But, what if we refactor to

```
data NatPred2' nat_arrow_bool = MkNP nat_arrow_bool
type NatPred2 = NatPred2' (Nat -> Bool)
```

The two `NatPred`s should behave identically in all settings, except that `NatPred2` will not promote, because type synonyms do not promote. But, if we expand type synonyms before promotion, then `NatPred2` becomes `NatPred2' (Nat -> Bool)`, which is perfectly promotable. And, tantalizingly, it promotes _correctly_, to `NatPred2' (TyFun Nat Bool -> *)`. Huzzah!

So, the implementation plan here is:
1. Detect datatypes that have members that do not promote but are fixable. This includes arrow members, and members whose types are type synonyms. (But see below for ways of expanding this set.)
2. Refactor the datatype to make the unpromotable bit a parameter to the datatype.
3. Make a type synonym to make this refactoring transparent to term-level users.
4. Promote as usual, making sure always to expand type synonyms.

What's unusual about this approach is that it means that the code the users write is not exactly what they get, at the term level. Currently, in a `promote` or `singletons` call, the user's code informs the promoted and singletonized versions but is otherwise passed through unchanged. This new idea would change that, but in (what should be) a harmless way.

Future expansion ideas: we could do something similar with `String`, promoting to `Symbol`. Functions that treat `String` like a list won't work, but just storing `String`s would work fine. And, we could also think about doing this trick to get, say, `Integer` to promote to some custom type-level `Integer` kind.

RyanGlScott:
#450 allows "semi-automatic" promotion of `Natural` to `Nat` etc. by allowing you to specify special name mappings during promotion. It doesn't implement the idea of sharing the same underlying data type (as in `NatPred2'` above), but the overall idea is very similar.

PR #283, opened 2018-01-02
[WIP] Try proof of concept for replacing arrow types with ty params

imalsogreg:
I've hacked up an implementation of the idea in https://github.com/goldfirere/singletons/issues/82 for promoting data decls with constructors that contain arrows.

For example:

```
*Data.Singletons T S S Data.Singletons.Prelude Data.Singletons Language.Haskell.TH.Ppr Language.Haskell.TH.Desugar
λ> T.runQ ([d| data Foo a = Foo { myfn :: Int -> Int } |] >>= desugar  >>= mangleDeclsWithTypes) >>= putStrLn . pprint . sweeten
data Foo'_0 a_1 tmp_rec_myfn = Foo_2 {myfn_3 :: tmp_rec_myfn}
type Foo_4 a_1 = Foo' (GHC.Types.Int -> GHC.Types.Int)
```

Apologies for not cleaning any of it up yet.

I am a little stuck on a conceptual problem - I have  a test case that derives the new data and type declarations in another project, basically:

```
$(singletons [d|
  data Foo = Bar { myFun :: Int -> Int }
  |])
```

When I compile this, I am hitting the TH phase restriction, I think:

```
[2 of 2] Compiling Testing          ( src/Testing.hs, interpreted )

src/Testing.hs:16:3: error:
    Looking up Bar' in the list of available declarations failed.
This lookup fails if the declaration referenced was made in the same Template
Haskell splice as the use of the declaration. If this is the case, put
the reference to the declaration in a new splice.
   |
16 | $(singletons [d|
   |   ^^^^^^^^^^^^^^...
Failed, one module loaded.
```

I'm wondering if there is a workaround for this? If the issue is too unclear due to the large number of poorly-named functions in my branch, or the large amount of commented-out garbage, I can bump the thread after cleaning those up :) Thanks for any clues that could get me unstuck!

### DIFF
--- a/src/Data/Singletons/Internal.hs
+++ b/src/Data/Singletons/Internal.hs
@@ -1,9 +1,9 @@
 {-# LANGUAGE MagicHash, RankNTypes, PolyKinds, GADTs, DataKinds,
-             FlexibleContexts, FlexibleInstances,
+             FlexibleContexts, FlexibleInstances, LambdaCase,
              TypeFamilies, TypeOperators, TypeFamilyDependencies,
              UndecidableInstances, TypeInType, ConstraintKinds,
              ScopedTypeVariables, TypeApplications, AllowAmbiguousTypes,
-             PatternSynonyms, ViewPatterns #-}
+             PatternSynonyms, TupleSections, ViewPatterns #-}
 
 -----------------------------------------------------------------------------
 -- |
@@ -24,12 +24,20 @@
 module Data.Singletons.Internal (
     module Data.Singletons.Internal
   , Proxy(..)
+  , wrap
   ) where
 
+import Data.Bool
+import Data.Either
+import Data.List (foldl')
+import Data.Maybe
 import Data.Kind
 import Unsafe.Coerce
 import Data.Proxy ( Proxy(..) )
 import GHC.Exts ( Proxy#, Constraint )
+import qualified Language.Haskell.TH as T
+import qualified Language.Haskell.TH.Syntax as T
+import qualified Language.Haskell.TH.Desugar as D
 
 -- | Convenient synonym to refer to the kind of a type variable:
 -- @type KindOf (a :: k) = k@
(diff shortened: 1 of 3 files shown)


RyanGlScott:
Hi @imalsogreg. I'd like to review this, but before we do that, I need a better understanding of what exactly this PR is trying to accomplish. A write-up that describes what is going on would be super helpful. Some questions that popped into my mind are:

* What happens in the presence of multiple arrows? e.g., `data Foo = MkFoo (Bool -> Ordering -> Bool)`
* In the example you posted above, you have `type Foo = Foo' (Int -> Int)`. Will there also be a promoted version like `type PFoo = Foo' (Int ~> Int)`?
* Where do the extra type variables get placed in a data type with other type parameters? For instance, if you have `newtype Endo a = Endo (a -> a)`, which of the following would be generated?
  * `newtype Endo' a tmp = ...`
  * `newtype Endo' tmp a = ...`
* Will this hack also apply to `Symbol` and `Nat`?

imalsogreg:
Thanks @RyanGlScott! I have in mind a specific type in a library which I would like to wrap in a `singletons` quasi quote:
```
data GraphKL a = GraphKL {
    toGraphKL    :: KL.Graph,
    fromVertexKL :: KL.Vertex -> a,
    toVertexKL   :: a -> Maybe KL.Vertex
    }
```
...actually I'm naively trying to wrap the entirety of the `alga` library https://github.com/snowleopard/alga in order to make use of graphs at the type level. A compiler error alerted me to the limitation in `singletons` that prevents lifting of data types that contain functions - with that error in mind I got more info about this limitation from the singletons paper. In the second argument to `GraphKL` above, the attempt to lift `KL.Vertex -> a` fails, as does `a -> Maybe KL.Vertex`.

A possible solution was given in https://github.com/goldfirere/singletons/issues/82 : let `singletons` refactor a data declaration with function-typed fields in its constructors into a pair:

 - a new data declaration just like the original but where all contained functions are replaced with fresh type parameters
 - a new type alias that applies the new data declaration at the right function types to recover the original data declaration

This refactoring applied to the type above would produce:
```
data GraphKL' a tmp1 tmp2 = GraphKL {
    toGraphKL    :: KL.Graph,
    fromVertexKL :: tmp1,
    toVertexKL   :: tmp2
    }

type GraphKL a = GraphKL' a (KL.Vertex -> a) (a -> Maybe KL.Vertex)
```

(After this point, my shallow understanding of `singletons` starts to get in the way). The resulting `GraphKL' a (KL.Vertex -> a) (a -> Maybe KL.Vertex)` promotes to `GraphKL' a (TyFun KL.Vertex a -> *)` (This claim comes from the issue #82 writeup. I don't understand how it is justified). With the transparent refactoring, the user gets a lifted version of their data type (presumably with the contained functions defunctionalized), and they get back also value-level code that is behaviorally similar to the code they wrote in the splice. (This 'behaviorally similar' code (the new data declaration applied to the proper function-typed arguments to recover the original data declaration) differs from the `singletons`ization of normal quoted data declarations - normal ones produce lifted versions and pass the unlifted code through untouched).

 * *What happens with multiple arrows?* - I think `type Foo = Foo' (Bool -> (Ordering -> Bool))` would be produced by my implementation, and how this gets lifted to the kind level, I'm not sure - need to reread the paper.
 * *Will `type Foo = Foo' (Int -> Int)` be lifted to `PFoo = Foo' (Int -> Int)`* - I'm tempted to say no, type aliases don't get lifted at all. But that leaves me confused about how this refactoring scheme results in useful lifted data types in the first place.
 * *Where do added type parameters sit relative to user-written type variables* - Generated type variables go at the *head* of the list of binders in the generated data declaration, so that application of the derived type alias consumes them and leaves the user-defined ones available. I believe my implementation gets this wrong and sticks them at the *tail*. We want this behavior:
   * `data Foo a = Foo (a -> Int)`
   *  (singletons refactor)
   * `data Foo' b a = Foo b`
   * `type Foo a = Foo' (a -> Int)`
 * *Will this hack apply to `Symbol` and `Nat`* - I believe so, but my reasoning here is shaky.

Thanks a lot for the probing questions! Really helped me to see what I have grokked so far and what I have not.

RyanGlScott:
I have to say, this is pretty ambitious! :) But I still feel like there are some details to be fleshed out here, so be patient if I keep asking questions about the design.

I understand the intent of this hack, but the part that's unclear to me is how, algorithmically, it will be applied. You've given two examples of what the end result would look like, which is great, but I'm still not sure how exactly one arrives at those end results.

Let's look at an example of what happens with a "normal" data type. This:

```haskell
$(singletons [d| data Foo = MkFoo Bool |])
```

Is turned into this:

```haskell
data Foo = MkFoo Bool

data instance Sing (z :: Foo) where
  SMkFoo :: Sing b -> Sing (MkFoo b)

instance SingKind Foo where
  fromSing (SMkFoo b) = fromSing b
  toSing (MkFoo b) = withSomeSing (toSing b) $ SomeSing . SMkFoo
```

So far, so good. Now what happens if we change the definition of `Foo` to this?

```haskell
$(singletons [d| data Foo = MkFoo (Bool -> Bool) |])
```

We cannot directly promote `Bool -> Bool`, so `singletons` needs to apply some sort of transformation to make this work. I'll attempt to give rule describing what happens:

* Generate a fresh data type with the same name (but with an apostrophe appended at the end).
* When a field whose type is of the form `Foo -> Bar` is encountered, replace that field with a fresh type variable `tmp_xyz` (where `xyz` is an appropriately fresh number). Append `tmp_xyz` to the list of type variables for this new data type.
* When all of this is done, generate the following two type synonyms:

  ```haskell
  type Foo = Foo' (Bool -> Bool)
  type PFoo = Foo' (Bool ~> Bool)
   ```

Notice that there's two `type`s here: `Foo`, the unpromoted (original) type, and `PFoo`, the promoted version, complete with a promoted function arrow `(~>)`. Why do we need `PFoo`? It's because of the `Sing` and `SingKind` instances that must follow:

```haskell
data instance Sing (z :: PFoo) where
  SMkFoo :: Sing b -> Sing (MkFoo b :: PFoo)

instance SingKind PFoo where
  fromSing (SMkFoo b) = fromSing b
  toSing (MkFoo b) = withSomeSing (toSing b) $ SomeSing . SMkFoo
```

That `SingKind` instance won't typecheck unless we use `PFoo`, since it relies on the `SingKind` instance for `(~>)`! This is why I'm making such a big deal out of this point.

This is also why I asked about the double-arrow case of, say, `data Foo = MkFoo (Bool -> (Bool -> Bool))`. If we blindly apply the "algorithm" above, we would get:

```haskell
data Foo' tmp = MkFoo tmp
type Foo = Foo' (Bool -> (Bool -> Bool))
type PFoo = Foo' (Bool ~> (Bool -> Bool))
```

Note that `PFoo` uses both `(~>)` and `(->)`, since it stops when it encounters the first arrow type! I don't think this is what we want. This leads me to believe that what we actually want is something like:

```haskell
data Foo' tmp_1 tmp_2 = MkFoo (tmp_1 Bool (tmp_2 Bool Bool))
type Foo = Foo' (->) (->)
type PFoo = Foo' (~>) (~>)
```

That way, we convert every occurrence of `(->)` to `(~>)` when promoted. I also think this work better if we also apply an analogous hack for `Nat` and `Symbol`, since we can simply have a list of type constructors that we map to other type constructors (i.e., map `Text`, `Natural`, and `(->)` to `Symbol`, `Nat`, and `(~>)`, respectively). I haven't implemented this, of course, so I might be wrong here, but that's what my intuition leads me to believe.

Does all this make sense?

Once we've nailed down the algorithm, I'd request that you put this commentary in the comments of the code somewhere so that one can get caught up to speed more easily in the future.

goldfirere:
I haven't looked at any of the code -- still working on design.

I'm not so worried about the `data Foo = MkFoo (Bool -> Bool -> Bool)` case. You would generate `type Foo = Foo' (Bool -> Bool -> Bool)` and `type PFoo = Foo' (Bool ~> Bool ~> Bool)`, where the latter argument is gotten by simply promoting `Bool -> Bool -> Bool`. I don't see why this case is any harder than the `Bool -> Bool` case, if you use `promoteType`.

The idea of abstracting over the arrow is very intriguing... but unfortunately a no-go. The promoted arrow `(~>)` can't be partially applied. Having written that, I realize that `singletons` solves the partial application problem.... so maybe we could do this:

```
data Foo a b
  = MkFoo1 (Bool -> Bool -> Bool)
  | MkFoo2 (Maybe a -> Maybe b)
```

and transform it to

```
data Foo' arr a b
  = MkFoo1 (arr @@ Bool @@ (arr @@ Bool @@ Bool))
  | MkFoo2 (arr @@ Maybe a @@ Maybe b)

type Foo = Foo' (TyCon2 (->))
type PFoo = Foo' (:~>$)
```

That actually seems quite appealing, and much better than what I proposed in #82. It will really clutter, say, `:info` output, though. Yet there's something elegant about it. (Note that `singletons` doesn't currently include `(:~>$)` as a defunctionalization symbol for `(~>)`. But it really should. `genDefunSymbols` was able to make it without hurting itself.)

Doing this will open the question of how to promote type names. Currently, `Foo` promotes to `Foo` (see line 51 of `Data.Singletons.Promote.Type`). With this change, we'll need to promote `Foo` to `PFoo`. But how do we know to do this for `Foo` but not for types that don't undergo this transformation? I suppose we could look to see if `Foo` is a type synonym. But that might be hard to do. So we might end up needing to make `type PBar = Bar` synonyms for all types, even ones that don't undergo this transformation. That's not terrible, but just another wrinkle to be aware of. It *is* possible we'll be able to do this more intelligently (by knowing that `Foo` is special somehow), but I'm just not sure.

In answer to a question posed above somewhere: yes, this should work just fine with `Nat` and `Symbol`.

RyanGlScott:
> (Note that `singletons` doesn't currently include `(:~>$)` as a defunctionalization symbol for `(~>)`. But it really should. `genDefunSymbols` was able to make it without hurting itself.)

[This is the case](https://github.com/goldfirere/singletons/blob/master/src/Data/Singletons.hs#L136) in `singletons` HEAD.

RyanGlScott:
I'm going to close this, since this PR has been inactive for over two years and the `master`  branch has diverged significantly in that time. Please do submit a new PR if you would like to see this through.

Issue #225, opened 2017-07-10
Figure out a way to cache the Travis build


RyanGlScott:
Currently, the Travis CI job for `singletons` rebuilds all of its dependencies, every time. This makes me immensely sad.

Unfortunately, actually making Travis cache the dependencies is easier said than done. The thorn in our side is the need to install `th-desugar` from a submodule separately from all other dependencies. The usual Travis script that caches Hackage dependencies, however, calculates whether dependencies have changed or not with this command:

```bash
$ cabal install --only-dependencies --enable-tests --enable-benchmarks --dry -v > installplan.txt
```

You can't just throw `th-desugar` in there, since it doesn't play nicely with `--only-dependencies` and `--enable-tests`.

The proper solution would be to switch over to a `.travis.yml` file that uses `cabal new-build` (i.e., the approach [here](https://github.com/hvr/multi-ghc-travis/blob/800980d76f7a74f3cdfd76b3dff351d52d2c84ee/.travis.yml)). But migrating over to `cabal new-build` comes with its own set of problems, as the `singletons` test suite is designed with some "old-build" assumptions in mind. Namely,

1. It assumes that a file named `"../dist/build/autogen/cabal_macros.h"` exists. This assumption doesn't hold under `cabal new-build`.
2. It assumes that you're either using the global package database or one from a sandbox. `cabal new-build` organizes its own package databases, however.

@hvr had some ideas for porting the test suite over. Some possibilities:

1. Use a `Setup.hs` script to determine the right paths for `cabal_macros.h`/package databases. (I'm not especially eager to do this, since `Setup.hs` scripts come with their own problems.)
2. Augment `Paths_singletons` with the necessary information. This would require waiting for some number of `Cabal` releases, but it would likely be a clean solution.

ezyang:
I think if you take an approach similar to how cabal-testsuite goes about doing it, you might be able to handle the "invoke GHC to run tests" pattern (because this is exactly what cabal-testsuite does.) The key is that using the Cabal library and the path to the `dist` directory (which can be passed by hand or inferred from the path of the test executable), you can pull out all of the other information you need to do it. In cabal-testsuite, it loads in via https://github.com/haskell/cabal/blob/master/cabal-testsuite/Test/Cabal/Monad.hs (`--builddir` into `testArgDistDir`), and then we load the config with `getPersistBuildConfig` and then use values from the `lbi` to do the rest of our business. Of particular interest is the `withPackageDB` field. Let me know if you have other questions.

[...]

PR #284, opened 2018-01-07
WIP: GHC 8.4 (singletons-2.4)

### LINKED ISSUE #254: GHC 8.4 checklist
GHC 8.4 adds some extra goodies that would be nice to incorporate into `singletons`. This issue serves as a place to record all of these features as a reminder to fold them in before cutting an 8.4-compatible release.

* [ ] There are now [built-in `Div` and `Mod` type families](http://git.haskell.org/ghc.git/commit/fa8035e3ee83aff5a20fc5e7e2697bac1686d6a6) in `GHC.TypeNats` that use efficient machine operations under the hood. We can use these to vastly speed up the current `PShow` instance for `Nat`.

    While we're in town, we ought to work `singletons`' magic on `Log2` as well.
* [ ] Due to [recent work on data families](https://ghc.haskell.org/trac/ghc/ticket/12369), it is now possible to write `Sing` without extraneous type variables, like so:

  ```haskell
  data family Sing :: k -> Type
  data instance Sing :: [a] -> Type where ...
  ```

  As opposed to before, where we had to conjure up unused type variables to work around GHC restrictions:

  ```haskell
  data family Sing (a :: k)
  data instance Sing (z :: [a]) where ...
  ```

  In order to apply this change to its fullest, we will first need to update `th-desugar` to allow GADTs with explicit return kinds. See https://github.com/goldfirere/th-desugar/issues/68.

RyanGlScott:
This provides the necessary changes to the `singletons` library to support GHC 8.4. They're fairly minimal this go around. They include:

* Adding `MonadIO` instances for some newtypes, since `MonadIO` is now a superclass of `Quasi`.
* Implementing `qAddCorePlugin` for `Quasi` instances.
* Turning instances of the form `data instance Sing (z :: Foo)` to `data instance Sing :: Foo -> Type` where possible (see #254).
* Using the built-in type families `Div`, `Mod`, and `Log2` from `GHC.TypeNats` within `Data.Singletons.TypeLits` (see #254).

Fixes #254.

There's one remain snag: the Travis build won't currently work, since not all dependencies from Hackage currently build with GHC 8.4.1. We could conceivable work around this with `head.hackage`, but it would require a non-trivial amount of effort to make `singletons`' test suite support `cabal new-build` (see #225).


src/Data/Singletons/TypeLits.hs
```diff
@@ -98,32 +100,63 @@ no_term_level_syms = error "The kind `Symbol` may not be used at the term level.
...
-
-  divMod' :: Nat -> Nat -> Nat -> Nat -> (Nat, Nat)
-  divMod' 0 _ q u = (q, u)
-  divMod' n y q 0 = divMod' (n-1) y (q+1) y
-  divMod' n y q u = divMod' (n-1) y q     (u-1)
-
-  div :: Nat -> Nat -> Nat
-  div _ 0 = error "Division by zero."
-  div x y = fst (divMod x y)
-  infixl 7 `div`
```
RyanGlScott: One question you might ask is: do the built-in `Div` and `Mod` type families from `GHC.TypeNats` have the same fixity (`infixl 7`)? Sadly, the answer is currently "no", although I'm seeking to correct this discrepancy in [GHC Trac #14640](https://ghc.haskell.org/trac/ghc/ticket/14640).

CHANGES.md
```diff
@@ -1,8 +1,8 @@
 Changelog for singletons project
 ================================
 
-next
-----
+2.4
```
goldfirere: This should mention that we're switching support to 8.4.

singletons.cabal
```diff
@@ -42,7 +42,7 @@ source-repository this
 
 library
   hs-source-dirs:     src
-  build-depends:      base >= 4.10 && < 5,
+  build-depends:      base >= 4.11 && < 5,
```
goldfirere: Should we have 4.12 as an upper bound?

RyanGlScott: Good idea, fixed.

goldfirere:
To be clear, please merge when ready.

PR #288, opened 2018-01-14
Fix #287 by preferring `a ~> b` over `TyFun a b -> Type`

### LINKED ISSUE #287: Can't promote or single an instance for (a -> b)
While working on https://github.com/goldfirere/singletons/issues/184#issuecomment-355857980, I ran into an issue: the following instance will not single:

```haskell
{-# LANGUAGE DefaultSignatures #-}
{-# LANGUAGE GADTs #-}
{-# LANGUAGE InstanceSigs #-}
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE TypeInType #-}
module Bug where

import Prelude ()
import Data.Singletons.TH

$(singletons [d|
  class Semigroup a where
    (<>) :: a -> a -> a

  instance Semigroup b => Semigroup (a -> b) where
    f <> g = \x -> f x <> g x
  |])
```

```
$ /opt/ghc/8.4.1/bin/ghc Bug.hs
[1 of 1] Compiling Bug              ( Bug.hs, Bug.o )

Bug.hs:16:3: error:
    • Expected kind ‘TyFun a b -> *’,
        but ‘Apply (Apply TFHelper_6989586621679033749Sym0 a_a6en) a_a6eo’ has kind ‘TyFun
                                                                                       (TyFun a b) *
                                                                                     -> *’
    • In the type ‘Apply (Apply TFHelper_6989586621679033749Sym0 a_a6en) a_a6eo’
      In the type instance declaration for ‘<>’
      In the instance declaration for
        ‘PSemigroup (TyFun a_a6dJ b_a6dI -> GHC.Types.Type)’
   |
16 | $(singletons [d|
   |   ^^^^^^^^^^^^^^...
```

Moreover, I can't even find an alternative way to write the instance that does work. This also fails with the same error:

```haskell
  instance Semigroup b => Semigroup (a -> b) where
    f <> g = h
      where
        h x = f x <> g x
```

This also fails, but with a different error:

```haskell
  instance Semigroup b => Semigroup (a -> b) where
    (f <> g) x = f x <> g x
```
```
$ /opt/ghc/8.4.1/bin/ghc Bug.hs
[1 of 1] Compiling Bug              ( Bug.hs, Bug.o )

Bug.hs:16:3: error:
    Function being promoted to (<>@#@$) has too many arguments.
   |
16 | $(singletons [d|
   |   ^^^^^^^^^^^^^^...

Bug.hs:16:3: error: Q monad failure
   |
16 | $(singletons [d|
   |   ^^^^^^^^^^^^^^...
```

Promotion doesn't work either. If I try to promote either of the first two variants of the program, such as:

```haskell
$(promote [d|
  class Semigroup a where
    (<>) :: a -> a -> a

  instance Semigroup b => Semigroup (a -> b) where
    f <> g = \x -> f x <> g x
  |])
```

It fails with:

```
$ /opt/ghc/8.4.1/bin/ghc Bug.hs
[1 of 1] Compiling Bug              ( Bug.hs, Bug.o )

Bug.hs:16:3: error:
    • Expected kind ‘TyFun a b -> *’,
        but ‘Apply (Apply TFHelper_6989586621679033837Sym0 a_a6fN) a_a6fO’ has kind ‘TyFun
                                                                                       (TyFun a b) *
                                                                                     -> *’
    • In the type ‘Apply (Apply TFHelper_6989586621679033837Sym0 a_a6fN) a_a6fO’
      In the type instance declaration for ‘<>’
      In the instance declaration for
        ‘PSemigroup (TyFun a_a6f9 b_a6f8 -> GHC.Types.Type)’
   |
16 | $(promote [d|
   |   ^^^^^^^^^^^...
```

And attempting to promote this:

```haskell
$(promote [d|
  class Semigroup a where
    (<>) :: a -> a -> a

  instance Semigroup b => Semigroup (a -> b) where
    (f <> g) x = f x <> g x
  |])
```

fails with:

```
$ /opt/ghc/8.4.1/bin/ghc Bug.hs
[1 of 1] Compiling Bug              ( Bug.hs, Bug.o )

Bug.hs:16:3: error:
    • Number of parameters must match family declaration; expected 2
    • In the type family declaration for ‘TFHelper_6989586621679033740’
   |
16 | $(promote [d|
   |   ^^^^^^^^^^^...
```

I have no choice but to define these instances by hand as a result.

RyanGlScott:
Currently, `singletons` always generated promoted function arrows as `TyFun a b -> Type`. However, this greatly confuses `defunctionalize`, as doesn't distinguish between an occurrence of `(->)` in `TyFun a b -> Type` from other occurrences of `(->)`. This leads to the shenanigans discovered in #287.

This patch avoids the issue entirely by generating `a ~> b` instead of `TyFun a b -> Type` wherever possible. I say "wherever possible" since there are some uses of `TyFun` that are currently unavoidable, such as when one generates a defunctionalization symbol of the form:

```haskell
type family Foo (a :: Bool) :: Bool

data IdSym0 (l :: TyFun Bool Bool) :: Type
```

Really, we ought to be generating `data IdSym0 :: Bool ~> Bool` there, but `th-desugar` currently lacks the ability to do this (see https://github.com/goldfirere/th-desugar/issues/68 for details). Luckily, the uses of `TyFun` that I _was_ able to remove were sufficient to fix #287, which is what counts in the end.

### DIFF
--- a/src/Data/Singletons/Promote/Defun.hs
+++ b/src/Data/Singletons/Promote/Defun.hs
@@ -158,9 +158,24 @@ defunctionalize name m_arg_kinds' m_res_kind' = do
 
           mk_rhs' ns  = foldType (DConT data_name) (map DVarT ns)
 
-      decls <- go (n - 1) m_args (addStar_maybe (buildTyFun_maybe m_arg m_result)) mk_rhs'
+      decls <- go (n - 1) m_args (buildTyFunArrow_maybe m_arg m_result) mk_rhs'
       return $ suppress : data_decl : app_decl : decls
 
+-- Shorthand for building (k1 ~> k2)
+buildTyFunArrow :: DKind -> DKind -> DKind
+buildTyFunArrow k1 k2 = DConT tyFunArrowName `DAppT` k1 `DAppT` k2
+
+buildTyFunArrow_maybe :: Maybe DKind -> Maybe DKind -> Maybe DKind
+buildTyFunArrow_maybe m_k1 m_k2 = do
+  k1 <- m_k1
+  k2 <- m_k2
+  return $ DConT tyFunArrowName `DAppT` k1 `DAppT` k2
+
+-- Shorthand for building (TyFun k1 k2 ~> Type)
+--
+-- We'd like to replace all uses of this with `buildTyFunArrow`, but we can't
+-- do this until https://github.com/goldfirere/th-desugar/issues/68
+-- is implemented.
 buildTyFun :: DKind -> DKind -> DKind
 buildTyFun k1 k2 = DConT tyFunName `DAppT` k1 `DAppT` k2
 
@@ -168,7 +183,7 @@ buildTyFun_maybe :: Maybe DKind -> Maybe DKind -> Maybe DKind
 buildTyFun_maybe m_k1 m_k2 = do
   k1 <- m_k1
   k2 <- m_k2
-  return $ DConT tyFunName `DAppT` k1 `DAppT` k2
+  return $ buildTyFun k1 k2
 
 -- Counts the arity of type level function represented with TyFun constructors
 tyFunArity :: DKind -> Int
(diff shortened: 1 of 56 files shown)


src/Data/Singletons/Promote/Defun.hs
```diff
@@ -177,18 +192,11 @@ tyFunArity (DArrowT `DAppT` (DConT tyFunNm `DAppT` _ `DAppT` b) `DAppT` DStarT)
   = 1 + tyFunArity b
 tyFunArity _ = 0
 
--- Checks if type is (TyFun a b -> *)
-isTyFun :: DKind -> Bool
-isTyFun (DArrowT `DAppT` (DConT tyFunNm `DAppT` _ `DAppT` _) `DAppT` DStarT)
-  | tyFunName == tyFunNm
-  = True
-isTyFun _ = False
```
RyanGlScott: Amusingly, there appears to be an `isTyFun` function that may have been used at one point to distinguish the `(->)` in `TyFun a b -> Type` from other uses of `(->)`. But at the moment, it's not used anywhere in `singletons`, and with this patch, it becomes unnecessary.

PR #290, opened 2018-01-20
Use cabal new-build for testing

### LINKED ISSUE #225: Figure out a way to cache the Travis build

RyanGlScott:
This finally allows us to test `singletons` on GHC 8.4.1 and HEAD. The new way to invoke tests is now:

```
# Assuming you have ghc-8.4.1 and cabal-install-2.1 or later on your PATH
cabal new-configure --enable-tests
cabal new-build
cabal new-test # Assuming you are in the directory containing singletons.cabal
```

This changes the test suite to invoke `cabal new-exec ghc -- ...`, which piggybacks on the previously built files in the store. A bit fiddly, but it does work.

Fixes #225.


RyanGlScott:
@goldfirere, this approach does require the use of a development version of `cabal-install` (slated to be released sometime alongside GHC 8.4), so I'm not sure if we should foist that requirement onto everyone running the tests. What do you think?

tests/compile-and-dump/Promote/GenDefunSymbols.ghc84.template
```diff
@@ -9,7 +9,7 @@ Promote/GenDefunSymbols.hs:0:0:: Splicing declarations
         = snd ((GHC.Tuple.(,) LiftMaybeSym1KindInference) GHC.Tuple.())
     data LiftMaybeSym1 (l :: TyFun a0123456789876543210 b0123456789876543210
                              -> Type) (l :: TyFun (Maybe a0123456789876543210) (Maybe b0123456789876543210))
-      = forall arg. Data.Singletons.Internal.SameKind (Apply (LiftMaybeSym1 l) arg) (LiftMaybeSym2 l arg) =>
+      = forall arg. singletons-2.4.1:Data.Singletons.Internal.SameKind (Apply (LiftMaybeSym1 l) arg) (LiftMaybeSym2 l arg) =>
```
goldfirere: Can we scrub the version number from these files before comparing? That seems more robust.

RyanGlScott:
For me, the most pressing issue is just being able to have tests and Haddocks successfully running on each commit. It's nice in theory to say "we'll just test before pushing", but sometimes we forget to check! (Plus, if other folks make pull requests, it'd be nice to check their work too.)

I'll let you make the final call on whether to merge now or later.

tests/compile-and-dump/Promote/GenDefunSymbols.ghc84.template
RyanGlScott: Oh dear, I somehow missed that. Yes, we should absolutely scrub versions out using the `sed` machinery.

goldfirere:
Well, I suppose the package only works with a pre-release GHC, so I suppose it's OK if it also only works with a pre-release cabal. Go for it.

Issue #184, opened 2017-05-31
Support monads


goldfirere:
We're surprisingly close to being able to support monads (and, with them, list comprehensions). Singletonizing the basic class definitions works. But there's danger on the horizon.

The problem has all to do with CUSKs. As `Note [CUSKification]` in Promote describes, we have little control over when GHC infers a CUSK. And this causes havoc when we want GHC to infer the kind of higher-kinded type variables. Specifically, the `PFunctor` class has a CUSK, and so the user needs to write `class Functor (f :: * -> *)`. Putting in the kind annotations here isn't so, so terrible.

What's worse is that we run into similar problems elsewhere. I haven't tracked down the precise details, but compiling Base.hs on the `monads` branch led to an error where a defunctionalization symbol has a CUSK, and thus inferring that a kind variable should be higher-kinded fails.

Short of fixing CUSKs in GHC, the solution might be to emit GADT-style declarations of defunctionalization symbols instead of H98-style. This would allow us finer control of whether we have a CUSK. (Recall that, with `TypeInType`, all kind variables must be explicitly quantified in order to have a CUSK. So we just won't do that.) This is tantalizingly close. But I'm not going to get there tonight.

Checklist:

* [x] Add `Functor`, `Applicative`, `Monad`, `Alternative`, `MonadPlus`, and `MonadZip` classes, as well as supporting functions. (https://github.com/goldfirere/singletons/pull/314)
* [x] Add `Identity`, `Const`, `Foldable`, and `Traversable`, as well as supporting functions. (https://github.com/goldfirere/singletons/pull/337)
* [ ] Add support for deriving `Functor`, `Foldable`, and `Traversable`. (#344)

RyanGlScott:
If we're adding an `Applicative` class as part of this work, then we'll need this instance:

```haskell
instance Monoid a => Applicative ((,) a)
```

But currently, we don't have a `Monoid` (or `Semigroup`) class within `singletons`. Adopting GHC 8.4 is good timing for this, now that `Semigroup` has become a superclass of `Monoid`, making adding the two in tandem much easier.

RyanGlScott:
Also, if we're going to be adding `Control.Monad` at some point, then we'll need to implement `Foldable`/`Traversable` as well, since it's required by the function:

```haskell
mapM :: (Traversable t, Monad m) => (a -> m b) -> t a -> m (t b)
```

Which is re-exported from `Control.Monad`.

[...]

PR #291, opened 2018-01-23
Add promoted/singled versions of Semigroup/Monoid

RyanGlScott:
Addresses one part of #184 (specifically, https://github.com/goldfirere/singletons/issues/184#issuecomment-355857980) by introducing promoted/singled versions of `Semigroup` and `Monoid`, as well as exporting them from the `Prelude` to match `base-4.11`. This also adds a `Semigroup`/`Monoid` instance for `Symbol` using `AppendSymbol`.

Unresolved question: do we want to single all of the newtype modifiers from `Data.Monoid` and `Data.Semigroup`? They're perhaps of dubious value (although one possible use case would be to add an indirect `Monoid` instance for `Nat` using [`Sum`](https://hackage.haskell.org/package/base-4.10.1.0/docs/Data-Monoid.html#t:Sum) and [`Product`](https://hackage.haskell.org/package/base-4.10.1.0/docs/Data-Monoid.html#t:Product)).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,14 @@ next
 * Add `(%<=?)`, a singled version of `(<=?)` from `GHC.TypeNats`, as well as
   defunctionalization symbols for `(<=?)`, to `Data.Singletons.TypeLits`.
 
+* Add `Data.{Promotion,Singletons}.Prelude.{Semigroup,Monoid}`, which define
+  promoted and singled versions of the `Semigroup` and `Monoid` type classes.
+
+  `Symbol` is now has promoted `Semigroup` and `Monoid` instances as well.
+  As a consequence, `Data.Singletons.TypeLits` no longer exports `(<>)` or
+  `(%<>)`, as they are superseded by the corresponding methods from
+  `PSemigroup` and `SSemigroup`.
+
 2.4.1
 -----
 * Restore the `TyCon1`, `TyCon2`, etc. types. It turns out that the new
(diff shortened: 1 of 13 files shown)


RyanGlScott:
Or, to put it another way, I didn't bother adding the newtype wrappers for `Data.Monoid` and `Data.Semigroup` since other modules in `singletons` don't appear to do this. (For instance, `Data.Singletons.Prelude.Ord` doesn't bother singling the [`Down`](https://hackage.haskell.org/package/base-4.10.1.0/docs/Data-Ord.html#t:Down) newtype).

goldfirere:
Looks good to me.

 * There's always a line between what to include and what not to. Use your best judgment. It would seem that singling those newtypes should be dead easy, but maybe I'm wrong about that.

 * Out of curiosity, was there any challenge in doing this? Or did just dropping the definitions of `Semigroup` and `Monoid` into a `singletonsOnly` just work? It's always fun when it just works.

Thanks!

RyanGlScott:
> It would seem that singling those newtypes should be dead easy, but maybe I'm wrong about that.

It likely would be. In fact, I'll look into doing that next (in a separate commit).

> Out of curiosity, was there any challenge in doing this? Or did just dropping the definitions of `Semigroup` and `Monoid` into a `singletonsOnly` just work?

Implementing this project is how I discovered #287, so there was one hurdle to overcome. But once that was fixed, the rest was smooth sailing.

PR #292, opened 2018-01-23
singleton analogue and defunctionalization symbols for <=?

mstksg:
This PR adds the singleton analogue (`%<=`) and defunctionalization symbols to fully cover the *GHC.TypeNats* API, partially just for the sake of completeness.

On a more practical note, though, a bunch of libraries built on native GHC.TypeLits/GHC.TypeNats still build their API's around `<=?`, instead of `CmpNat`, so they're incompatible with *singleton*'s `POrd`/`SOrd` typeclasses.  Having `%<=?` makes these libraries significantly easier to use.

### DIFF
--- a/src/Data/Singletons/TypeLits/Internal.hs
+++ b/src/Data/Singletons/TypeLits/Internal.hs
@@ -28,11 +28,13 @@ module Data.Singletons.TypeLits.Internal (
   Undefined, sUndefined,
   KnownNat, TN.natVal, KnownSymbol, symbolVal,
   type (^), (%^),
+  type (<=?), (%<=?),
   type (<>), (%<>),
 
   -- * Defunctionalization symbols
   ErrorSym0, ErrorSym1, UndefinedSym0,
   type (^@#@$),  type (^@#@$$),  type (^@#@$$$),
+  type (<=?@#@$),  type (<=?@#@$$),  type (<=?@#@$$$),
   type (<>@#@$), type (<>@#@$$), type (<>@#@$$$)
   ) where
 
@@ -169,7 +171,7 @@ $(genDefunSymbols [''Undefined])
 sUndefined :: a
 sUndefined = undefined
 
--- | The singleton analogue of '(TL.^)' for 'Nat's.
+-- | The singleton analogue of '(TN.^)' for 'Nat's.
 (%^) :: Sing a -> Sing b -> Sing (a ^ b)
 sa %^ sb =
   let a = fromSing sa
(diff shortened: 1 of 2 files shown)


RyanGlScott:
I was about to ask why you'd need this at all, given that there's already a `(%<=)` function in `SOrd`. But your practical note made me realize what the issue is: `(%<=)` uses the `CmpNat` type family internally, but `(%<=?)` uses the `(<=?)` type family. Alas, `CmpNat` and `(<=?)` are both wired-in type families, and `a <=? b` doesn't imply `CmpNat a b ~ EQ` for arbitrary `Nat`s `a` and `b`. Bummer.

It would be worth noting this subtle point in the Haddocks.

mstksg:
It is indeed a rather significant bummer.  I have a library I personally use just to get around this fact.  I'll add a note in the haddocks when I get the chance.

It is possible to write these functions in singletons, as well, but I'm not sure if it's within the scope of the library for things like

```
Sing (x :<= y) -> Sing (x :<=? y)
Sing (x :<=? y) -> Sing (x :<= y)
Sing (x :<=? y) -> Either (x :== y) (x :<= y)
```
etc.; I'm leaning on it more for being outside of the scope of *singletons*, but I'm not 100% sure.

RyanGlScott:
> I'm leaning on it more for being outside of the scope of *singletons*, but I'm not 100% sure.

I would agree with it being outside the scope of `singletons`.

mstksg:
Submitted note for review --

> Note that, because of historical reasons in GHC's `TN.Nat` API, `TN.<=?` is incompatible (unification-wise) with `<=` and the `PEq`, `SEq`, `POrd`, and `SOrd` instances for `Nat'`  `(a <=? b) ~ 'True` does not imply anything about `a <= b` or any other `PEq`/`POrd` relationships.
> This is provided for the sake of completeness and for compatibility with libraries with APIs built around `<=?`.  Newer codebases should use `CmpNat`, exposed through this library through the `POrd` and `SOrd` instances for `Nat`.

src/Data/Singletons/TypeLits/Internal.hs
```diff
@@ -183,6 +185,24 @@ infixr 8 %^
 -- Defunctionalization symbols for type-level (^)
 $(genDefunSymbols [''(^)])
 
+-- | The singleton analogue of 'TN.<=?' 
+--
+-- Note that, because of historical reasons in GHC's 'TN.Nat' API, 'TN.<=?'
+-- is incompatible (unification-wise) with '<=' and the 'PEq', 'SEq',
```
RyanGlScott: It's worth clarifying that when you refer to `(<=)` here, you're referring to the promoted `(<=)` family from `PEq` (and not the [`(<=)` from `GHC.TypeNats`](https://hackage.haskell.org/package/base-4.10.1.0/docs/GHC-TypeNats.html#t:-60--61-), since that's defined as `type (<=) x y = (x <=? y) ~ True`).

RyanGlScott:
Thanks!

PR #293, opened 2018-01-27
Add newtype modifiers to Data.Ord, Data.Semigroup, and Data.Monoid

RyanGlScott:
Addresses https://github.com/goldfirere/singletons/pull/291#issuecomment-360885936.

This patch looks larger than it actually is due to the need to reorganize the module hierarchy a bit to avoid import cycles.


goldfirere:
That business with `max_` and `min_` is regrettable but unavoidable. LGTM!

PR #295, opened 2018-01-28
Permit singling of expression and pattern signatures

### LINKED ISSUE #183: Singletonize pattern signatures
It would be nice if we could single

```
f (x :: Maybe Bool) = ...
```

or even

```
f (x :: Maybe a) = ...
```

where the latter binds a scoped type variable. Alas, this is not so easy, because of two features of GHC:

1. A scoped type variable must be just that, a variable. For example, consider this:

    ```
    g x = case Just x of
      (Just y :: Maybe Bool) -> y
    ```

    The natural singletonization is

    ```
    sG :: Sing x -> Sing (G x)
    sG sx = case SJust sx of
      (SJust sy :: Sing (blah :: Maybe Bool)) -> sy
    ```

    The problem is that GHC complains that `blah` should really be `Just xyz`. It won't *bind* `blah` to be `Just xyz` as a different design would do.

2. The alternative is to write the promoted version of the pattern instead of `blah`. But this doesn't work because GHC checks a pattern signature *before* checking the pattern itself. Consider

    ```
    f (Just a :: Maybe Bool) = "hi"
    ```

    If we singletonize to

    ```
    sF :: Sing x -> Sing (F x)
    sF (SJust sa :: Sing (Just a :: Maybe Bool)) = (sing :: Sing "hi")
    ```

    GHC complains that we don't yet know that the type `x` is `Just a`. We would need to do a pattern match without a signature and then do a `case` with a signature, to get the ordering right.

One escape hatch here is to use a `_` in place of `blah`. Of course, this requires `-XPartialTypeSignatures`, and even then, it emits a warning. We could ask users to disable this warning, but that just seems wrong. And no one is asking for this, anyway. So I give up for now. But if I ever return to this problem, I think the way forward is to do something like

```
sF x@(SJust sa) = case x of
  (_ :: Sing (Just a :: Maybe Bool)) -> sing :: Sing "hi"
```

as the singletonization of `f`, above. Yuck. But doable. And no yuckier than lots of other yuck in `singletons`.

RyanGlScott:
By promoting `DPat`s to a `ADPat`, a variant with extra type information, and changing `singPat` to generate extra information to use for creating an extra case expression, we can fix #183 with surprising ease.

### DIFF
--- a/src/Data/Singletons/Single.hs
+++ b/src/Data/Singletons/Single.hs
@@ -35,6 +35,7 @@ import qualified Data.Map.Strict as Map
 import Data.Map.Strict ( Map )
 import Data.Maybe
 import Control.Monad
+import Control.Monad.Writer
 import Data.List
 import qualified GHC.LanguageExtensions.Type as LangExt
 
@@ -478,7 +479,7 @@ singClause prom_fun num_arrows bound_names res_ki
     fail $ "Function being promoted to " ++ (pprint (typeToTH prom_fun)) ++
            " has too many arguments."
 
-  sPats <- mapM (singPat (Map.fromList var_proms)) pats
+  (sPats, sigPaExpsSigs) <- evalForPair $ mapM (singPat (Map.fromList var_proms)) pats
   sBody <- singExp exp res_ki
     -- when calling unSingFun, the promoted pats aren't in scope, so we use the
     -- bound_names instead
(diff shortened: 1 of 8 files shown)


tests/compile-and-dump/Singletons/T183.hs
```diff
@@ -0,0 +1,55 @@
...
+  -----
+  -- Other pattern features
+  -----
+
+  foo7 :: a -> b -> a
+  foo7 (x :: a) (_ :: b) = (x :: a)
+
+  foo8 :: forall a. Maybe a -> Maybe a
+  foo8 x@(Just (_ :: a) :: Maybe a) = x
+  -- foo8 x@(Nothing :: Maybe a)       = x
```
RyanGlScott: So it turns out that last line doesn't work at the moment, for a rather silly reason: it generates code like this:

```haskell
type family Let123X where
  Let123X = (NothingSym0 :: Maybe a)

type family Foo8 (x :: Maybe a) :: Maybe a where
  Foo8 (Nothing :: Maybe a) = Let123X

sFoo8 ::
  forall (t :: Maybe a).
  Sing t -> Sing (Foo8 t :: Maybe a)
sFoo8 SNothing
  = case SNothing of {
      (_ :: Sing (Nothing :: Maybe a))
        -> let
             sX :: Sing Let123X
             sX = SNothing :: Sing (NothingSym0 :: Maybe a)
           in sX }
```

Which doesn't typecheck:

```
../Bug.hs:14:3: error:
    • Could not deduce: a ~ a1
      from the context: t ~ 'Nothing
        bound by a pattern with constructor:
                   SNothing :: forall a. Sing 'Nothing,
                 in an equation for ‘sFoo8’
        at ../Bug.hs:(14,3)-(65,4)
      ‘a’ is a rigid type variable bound by
        the type signature for:
          sFoo8 :: forall a (t :: Maybe a). Sing t -> Sing (Apply Foo8Sym0 t)
        at ../Bug.hs:(14,3)-(65,4)
      ‘a1’ is a rigid type variable bound by
        the type signature for:
          sX :: forall a1. Sing Let6989586621679113062XSym0
        at ../Bug.hs:(14,3)-(65,4)
      Expected type: Sing Let6989586621679113062XSym0
        Actual type: Sing NothingSym0
    • In the expression: SNothing :: Sing (NothingSym0 :: Maybe a_aqR7)
      In an equation for ‘sX’:
          sX = SNothing :: Sing (NothingSym0 :: Maybe a_aqR7)
      In the expression:
        let
          sX :: Sing Let6989586621679113062XSym0
          sX = SNothing :: Sing (NothingSym0 :: Maybe a_aqR7)
        in sX
    • Relevant bindings include
        sX :: Sing Let6989586621679113062XSym0 (bound at ../Bug.hs:14:3)
        sFoo8 :: Sing t -> Sing (Apply Foo8Sym0 t)
          (bound at ../Bug.hs:14:3)
```

It turns out the culprit is the `sX :: Sing Let123X` type signature (commenting it out makes it typecheck). It looks like GHC can't figure out the `a` quantified by `sFoo8` should be the same `a1` that is an implicit argument to `Let123X`.

I would be that we could fix this by making that `a1` argument to `Let123X` visible and propagating that change through to its defunctionalization symbols, but I didn't have the energy to figure out how to do all of that in this patch. (It could probably be done separately, if need be.)

RyanGlScott: Actually, it turns out that this is an entirely orthogonal issue, as you can trigger this problem without using pattern signatures. See https://github.com/goldfirere/singletons/issues/296.

src/Data/Singletons/Single.hs
```diff
@@ -537,6 +557,82 @@ singPat _var_proms DWildPa = return DWildPa
...
+-- know that the type `z` is actually `Just x` until /after/ the entirety of
+-- the `SJust sX` pattern has been typechecked. (See Trac #12018 for an
+-- extended discussion on this topic.)
+--
+-- To work around this design, we resort to a somewhat unsightly trick:
+-- immediately after matching on all the patterns, we perform a case on every
+-- pattern with a pattern signature, like so:
+--
+--   sF :: forall (z :: Maybe a). Sing z -> Sing (F z)
+--   sF (SJust sX :: Sing z)
```
goldfirere: I'm worried about this approach. Whenever I do this trick, I do something like

```
f x@(T ...) = case x of (T ... :: ...) ->
```

Note the as-pattern. This means that GHC will always be convinced that the two scrutinees are the same. Here, you're doing it with type signatures, but I'm not sure that's always going to be enough. What if there is kind polymorphism? Ambiguity? Maybe I'm worried about nothing, but this doesn't smell right.

Of course, to make this work, we'd need as-patterns, which th-desugar doesn't support. So perhaps we should generate plain old TH.

tests/compile-and-dump/Singletons/T183.hs
goldfirere: This is perhaps related to the problem I'm worried about above.

src/Data/Singletons/Single.hs
RyanGlScott: On the contrary, I _do_ think this is always going to be enough. All that we're doing here is informing the typechecker that the `z` in `(SJust sX :: Sing z)` is equal to `(Just x :: Maybe a)`. The actual choice of expression that we scrutinize isn't important, since we take care to annotate both the original pattern and the scrutinized expression with `Sing z`, so there will never be ambiguity in that department.

goldfirere: Hm. That's a convincing response. OK. Proceed.

Issue #296, opened 2018-01-29
Failure to single let-bound subexpression with a type signature


RyanGlScott:
Originally observed in https://github.com/goldfirere/singletons/pull/295#discussion_r164288565. This program:

```haskell
import Data.Kind

data MyProxy (a :: Type) = MyProxy

f :: forall a. MyProxy a -> MyProxy a
f MyProxy = 
  let x = let z :: MyProxy a
              z = MyProxy in z
  in x
```

Will correctly promote, but not correctly single, as this program demonstrates:

```haskell
{-# LANGUAGE GADTs #-}
{-# LANGUAGE ScopedTypeVariables #-}
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE TypeApplications #-}
{-# LANGUAGE TypeInType #-}
{-# LANGUAGE UndecidableInstances #-}
module Bug where

import Data.Kind
import Data.Singletons.Prelude
import Data.Singletons.TH

$(singletons [d|
  data MyProxy (a :: Type) = MyProxy

  f :: forall a. MyProxy a -> MyProxy a
  f MyProxy =
    let x = let z :: MyProxy a
                z = MyProxy in z
    in x
  |])
```
```
$ /opt/ghc/8.4.1/bin/ghc Bug.hs -ddump-splices
[1 of 1] Compiling Bug              ( Bug.hs, Bug.o )
Bug.hs:(14,3)-(22,4): Splicing declarations
    singletons
      [d| f_a32l ::
            forall a_a3vI. MyProxy_a32j a_a3vI -> MyProxy_a32j a_a3vI
          f_a32l MyProxy_a32k
            = let
                x_a3vJ
                  = let
                      z_a3vK :: MyProxy_a32j a_a3vI
                      z_a3vK = MyProxy_a32k
                    in z_a3vK
              in x_a3vJ
          
          data MyProxy_a32j (a_a3vH :: Type) = MyProxy_a32k |]
  ======>
    data MyProxy_a6jb (a_a6jd :: Type) = MyProxy_a6jc
    f_a6ja :: forall a_a6j9. MyProxy_a6jb a_a6j9 -> MyProxy_a6jb a_a6j9
    f_a6ja MyProxy_a6jc
      = let
          x_a6je
            = let
                z_a6jf :: MyProxy_a6jb a_a6j9
                z_a6jf = MyProxy_a6jc
              in z_a6jf
        in x_a6je
[...]

PR #298, opened 2018-02-01
Fix #297 by explicitly binding kind variables in generated code

### LINKED ISSUE #297: How to adapt singletons to the "Treat kind and type vars identically with `forall`" proposal
If https://github.com/ghc-proposals/ghc-proposals/pull/103 is adopted (which it seems pretty likely that it will), then `singletons` will be in peril, as it quite often generates type signatures headed by `forall`s that also implicitly quantify kind variables. For example, this function's type signature:

```haskell
id :: a -> a
```

Is currently singled to:

```haskell
sId :: forall (t :: a). Sing t -> Sing (Apply IdSym0 t :: a)
```

Where the kind variable `a` is implicitly quantified, flying in the face of the `forall`-or-nothing rule of the future.

This `singType` function is probably quite relevant to this discussion:

https://github.com/goldfirere/singletons/blob/7f09b861db58f8e5d7b65b4886323fe5d5503957/src/Data/Singletons/Single/Type.hs#L19-L37

Notice how on line 35, the singled type signature's `forall` is constructed by simply zipping together the promoted argument names with their promoted types, which will certainly miss type variables that were promoted to kinds.

A tempting solution is to traverse the promoted types, extract all kind variables, and put them at the front of the `forall`. Sadly, this is not a strategy which will work in general. Consider this example:

```haskell
class Foo a where
  bar :: a -> b -> b
```

This currently singles to:

```haskell
class SFoo a where
  sBar :: forall (t1 :: a) (t2 :: b). Sing t1 -> Sing t2 -> Sing (Apply (Apply BarSym0 t1) t2 :: b)
```

However, if we simply extracted every kind variable we found from the promoted types and stuck them before the other `forall`'d variables, we'd end up with:

```haskell
class SFoo a where
  sBar :: forall a b (t1 :: a) (t2 :: b). Sing t1 -> Sing t2 -> Sing (Apply (Apply BarSym0 t1) t2 :: b)
```

Which would be incorrect: `a` was already quantified by the class declaration itself. So this algorithm would need to be aware of all the previously bound variables in scope.

RyanGlScott:
See `Note [Explicitly binding kind variables]`. In summary, we now track bound kind variables in `SgEnv`, and consult this to determine which kind variables to put an explicit `forall` in a singled type signature.

(Note that if this were to be merged, #295 would need to be updated slightly, as pattern signatures also have the ability to bind kind variables in singled definitions.)

### DIFF
--- a/src/Data/Singletons/Single/Monad.hs
+++ b/src/Data/Singletons/Single/Monad.hs
@@ -11,7 +11,8 @@ The SgM monad allows reading from a SgEnv environment and is wrapped around a Q.
 {-# LANGUAGE GeneralizedNewtypeDeriving, ParallelListComp, TemplateHaskell #-}
 
 module Data.Singletons.Single.Monad (
-  SgM, bindLets, lookupVarE, lookupConE,
+  SgM, bindLets, bindKindVars, allBoundKindVars,
+  lookupVarE, lookupConE,
   wrapSingFun, wrapUnSingFun,
   singM, singDecsM,
   emitDecs, emitDecsM
@@ -20,6 +21,8 @@ module Data.Singletons.Single.Monad (
 import Prelude hiding ( exp )
 import Data.Map ( Map )
 import qualified Data.Map as Map
+import Data.Set ( Set )
+import qualified Data.Set as Set
 import Data.Singletons.Promote.Monad ( emitDecs, emitDecsM )
 import Data.Singletons.Names
 import Data.Singletons.Util
@@ -33,13 +36,15 @@ import Control.Monad.Fail
 
 -- environment during singling
 data SgEnv =
-  SgEnv { sg_let_binds   :: Map Name DExp   -- from the *original* name
-        , sg_local_decls :: [Dec]
+  SgEnv { sg_let_binds       :: Map Name DExp   -- from the *original* name
+        , sg_bound_kind_vars :: Set Name -- See Note [Explicitly quantifying kinds variables]
+        , sg_local_decls     :: [Dec]
         }
 
 emptySgEnv :: SgEnv
-emptySgEnv = SgEnv { sg_let_binds   = Map.empty
-                   , sg_local_decls = []
+emptySgEnv = SgEnv { sg_let_binds       = Map.empty
+                   , sg_bound_kind_vars = Set.empty
+                   , sg_local_decls     = []
                    }
 
 -- the singling monad
@@ -90,6 +95,18 @@ bindLets :: [(Name, DExp)] -> SgM a -> SgM a
 bindLets lets1 =
   local (\env@(SgEnv { sg_let_binds = lets2 }) ->
                env { sg_let_binds = (Map.fromList lets1) `Map.union` lets2 })
+
+-- Add to the set of bound kind variables currently in scope.
+-- See Note [Explicitly binding kind variables]
+bindKindVars :: Set Name -> SgM a -> SgM a
+bindKindVars kvs1 =
+  local (\env@(SgEnv { sg_bound_kind_vars = kvs2 }) ->
+               env { sg_bound_kind_vars = kvs1 `Set.union` kvs2 })
+
+-- Look up the set of bound kind variables currently in scope.
+-- See Note [Explicitly binding kind variables]
+allBoundKindVars :: SgM (Set Name)
+allBoundKindVars = asks sg_bound_kind_vars
 
 lookupVarE :: Name -> SgM DExp
 lookupVarE = lookup_var_con singValName (DVarE . singValName)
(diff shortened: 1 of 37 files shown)


src/Data/Singletons/Single/Monad.hs
```diff
@@ -154,3 +171,66 @@ singDecsM :: DsMonad q => [Dec] -> SgM [DDec] -> q [DDec]
...
+
+We would first promote the types to kinds and generate an explicit forall
+mapping fresh type variables to the promoted types (step 1):
+
+  sIdentity :: forall (x :: a). Sing x -> Sing (Identity x)
+  sIdentity sX = sX
+
+Now, because we want to explicitly bind the kind variable `a`, we traverse
+the promoted types, extract the set of all kind variables, and put
+them at the front of the forall, like so (step 2):
```
goldfirere: In this case, we don't need a 2-step process. You have the `forall a` in the original code. So we know we're going to need to quantify the kind `a` in the singled code. However, in the event that the original code *doesn't* have an explicit quantification, we'd need something like this. So, either you track in-scope variables in the original source so that you can recreate GHC's implicit quantification, or you do it the way it's done here. To be honest, I slightly prefer the first way -- it's conceptually closer to the way implicit quantification works -- but this approach is fine too.

goldfirere: Actually, the first approach might be considerably simpler. You can do it all in the promotion pass, making the implicit quantification explicit in the `ADxxx` AST. Then, when singling, you just have to preserve explicit quantification. Still, I'm not asking you to rewrite this.

RyanGlScott: Hm, I think the Note may have been misleading, since you're describing I problem which I believe doesn't exist. You're worrying about what happens when `identity` is written as `identity :: a -> a`, but there's no difference in behavior from `singletons` end when `identity` is written with an explicit `forall` or not—either way, it'll be singled to `sIdentity` with an explicit `forall`. (Perhaps I'm misreading this comment, though...)

goldfirere: I think you're misreading the comment.

Suppose all quantification were explicit. Then, we could simply echo the quantification that the user wrote -- end of story. No need to look for/track free variables.

Of course, all quantification isn't explicit. But we can do a pre-pass over the AST and make it explicit, according to simple rules. Because we already have to take a pass over the AST during promotion, this explicit quantification could happen there.

Does that clarify?

RyanGlScott: > Of course, all quantification isn't explicit.

You keep saying this, but this simply isn't true in the context of `singletons` (and I'm talking about the current release, not just after this patch). When we single a definition,we _always_ explicitly quantify the variables, regardless of whether the original definition was explicitly quantified or not. As evidence, you can see for yourself that this definition:

```haskell
$(singletons [d| identity :: a -> a; identity x = x|])
```

Singles to:

```haskell
    sIdentity ::
      forall (t_a6BO :: a_a6BG).
      Sing t_a6BO -> Sing (Apply IdentitySym0 t_a6BO :: a_a6BG)
    sIdentity (sX :: Sing x_a6BN) = sX
```

In fact, you get the exact same singled definition when you explicitly quantify the `a` in the original source:

```haskell
$(singletons [d| identity :: forall a. a -> a; identity x = x|])
```

So I don't understand what the fuss is about.

goldfirere: I'm talking about explicit quantification in the original, user-written source, not in our output. If the original, user-written source had full explicit quantification, all those type variables become the kind variables we need to quantify over in our output. So we could the quantification in the original, user-written source to guide the quantification in the output.

RyanGlScott: > If the original, user-written source had full explicit quantification, all those type variables become the kind variables we need to quantify over in our output. So we could the quantification in the original, user-written source to guide the quantification in the output.

I mean, sure, that's one thing you could do. But I don't see that as making the implementation simpler. As you noted in https://github.com/goldfirere/singletons/pull/298#discussion_r165858547, the original source isn't always going to have explicit quantification, so we'd still need to do all of this free-variable calculation, but elsewhere in the code.

What's worse, storing the bound variables in `ADxxx` isn't always going to work, because not all of the types that we single are going to come from promotion. See this, for example:

https://github.com/goldfirere/singletons/blob/272f678dd83271a18d7ed3c181fe0f7b6395f3e8/src/Data/Singletons/Single.hs#L373-L380

where we have a scenario where we call `singType` on a type taken directly from `dsReify` (instead of through the promotion pipeline). If anything, we'd need even _more_ bookkeeping in order to propagate the right information through to code like this.

goldfirere: OK, if you say so. No further objections from me.

RyanGlScott:
I now recant my position in https://github.com/goldfirere/singletons/pull/298#discussion_r166021341—doing this in promotion is absolutely the way to go, for the reasons explained in https://github.com/goldfirere/singletons/issues/296#issuecomment-363823093. I've implemented this in #302, so I'll close this PR in favor of that one.

PR #302, opened 2018-02-10
Explicitly bind kinds during promotion

RyanGlScott:
See `Note [Explicitly binding kind variables]`. In summary, we now track bound kind variables in `PrEnv`, and stash the set of bound kind variables in scope in each `ALetDecEnv`, and use this information when determining which kind variables to explicitly quantify during singling.

This PR supersedes #298. Unlike #298, this does the bound variable tracking during promotion instead of singling. This results in slightly cleaner code in some cases (except for `singInstD`, but that function was a hot mess to being with) and will hopefully make it simpler to implement a fix for #296 when that becomes unblocked.

This is based on top of #295, to demonstrate the changes needed to `promotePat` in order to bind pattern signatures' kind variables.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,9 +3,12 @@ Changelog for singletons project
 
 2.5
 ---
-* `singletons` now generates `a ~> b` instead of `TyFun a b -> Type` whenever
-  possible. This may require enabling `TypeInType` in code which did not
-  previously need it.
+* Template Haskell-generated code may require `TypeInType` in scenarios which
+  did not previously require it:
+  * `singletons` now explicitly quantifies all kind variables used in explicit
+    `forall`s.
+  * `singletons` now generates `a ~> b` instead of `TyFun a b -> Type` whenever
+    possible.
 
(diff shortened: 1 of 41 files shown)


goldfirere:
I haven't gotten a chance to review this, but I'm OK forgoing that privilege. :) Do let me know if there's any particular spot that might want my attention.

Thanks.

RyanGlScott:
Ah yes, sorry for charging forth without a review! I wanted to land this since it would conflict with another pending pull request I'm working on.

The overall structure of this is a mirror of #298, but in `Data.Singletons.Promote`, which is why I decided to just go ahead and merge it. The one part of this that felt especially yucky was the changes I had to make to `singInstD`—namely, we have to recompute the bound variables there instead of stashing them in an `AInstDecl`. This is due to the fact that `AInstDecl`s don't store the type signatures for any of their methods, so you have to figure them out by reifying the class methods.

It's possible we might be able to refactor this some so that we could avoid this... do you think that's worth its own ticket?

Issue #74, opened 2014-05-14
Drop GHC 7.6 support


jstolarek:
Data.Singletons.Types is now redundant

PR #305, opened 2018-02-18
Adapt to th-desugar changes in #74

RyanGlScott:
This updates `singletons` to reflect the fairly significant `th-desugar` changes from https://github.com/goldfirere/th-desugar/pull/74. This patch looks massive, but really most of this is testsuite changes due to all generated data types now using GADT syntax. The changes weren't entirely trivial, though—here are the highlights:

1. Before, checking whether a data type used GADT syntax was dead simple—just check if a `DCon` had `Just ty` for its `Maybe DType` field (i.e., if it has an explicit return type). After this `th-desugar` patch, however, everything now uses GADT syntax, so this no longer works.

   To compensate, I introduced a `isProperGADT` function to `Data.Singletons.Partition`, which checks if a datatype is a _proper_ GADT that cannot be written using equivalent Haskell98 syntax, something which derived `Enum` instances disallow. However, this requires more than just a `Con` as an argument—it also requires information from the `DDataD` itself that isn't easily available when calling `mkEnumInstance`. I found that the cleanest solution was to add an extra `Bool` argument to `mkEnumInstance` which states whether a constructor comes from a proper GADT or not.

2. Switching generated defunctionalization data types over to GADT syntax proved to be surprisingly challenging, as my initial attempt ran into CUSK issues! Happily, I was able to sidestep this issue by dropping the kinds on the type variable binders of kind inference constructors. See the documentation for `dropTvbKind` for the full story.

3. Not surprisingly, I had to enable the `GADTs` extension in more places because, y'know, GADTs.

While I was in town, I implemented idea two from https://github.com/goldfirere/singletons/issues/254. Instead of generating this:

```haskell
data instance Sing (z :: Maybe a) where ...
```

We now generate this:

```haskell
data instance Sing :: Maybe a -> Type
```

Pretty nifty. There are some other things that one could be able to do with explicit return kinds, such as generating `data FooSym0 :: a ~> b` instead of `data FooSym0 (z :: TyFun a b)`, but these changes would not be as straightforward, so I'll tackle these separately.


tests/compile-and-dump/InsertionSort/InsertionSortImp.ghc84.template
```diff
@@ -7,11 +7,12 @@ InsertionSort/InsertionSortImp.hs:(0,0)-(0,0): Splicing declarations
...
-    data SuccSym0 (l :: TyFun Nat Nat)
-      = forall arg. SameKind (Apply SuccSym0 arg) (SuccSym1 arg) =>
-        SuccSym0KindInference
+    data SuccSym0 (l :: TyFun Nat Nat) :: Type
+      where
+        SuccSym0KindInference :: forall l arg.
+                                 SameKind (Apply SuccSym0 arg) (SuccSym1 arg) => SuccSym0 l
     type instance Apply SuccSym0 l = Succ l
-    data instance Sing (z :: Nat)
+    data instance Sing :: Nat -> Type :: Nat -> Type
```
RyanGlScott: If you're wondering why GHC is stuttering here with `:: Nat -> Type :: Nat -> Type`, it's due to [GHC Trac #14817](https://ghc.haskell.org/trac/ghc/ticket/14817), which I'm looking into now.

src/Data/Singletons/Deriving/Enum.hs
```diff
@@ -23,13 +23,11 @@ import Control.Monad
...
-mkEnumInstance mb_ctxt ty cons = do
+mkEnumInstance :: Quasi q => Bool -> Maybe DCxt -> DType -> [DCon] -> q UInstDecl
+mkEnumInstance is_gadt mb_ctxt ty cons = do
   when (null cons ||
-        any (\(DCon tvbs cxt _ f rty) -> or [ not $ null $ tysOfConFields f
-                                            , not $ null tvbs
-                                            , not $ null cxt
-                                            , isJust rty ]) cons) $
+        any (\(DCon _ _ _ f _) ->
+              is_gadt || or [ not $ null $ tysOfConFields f ]) cons) $
```
goldfirere: Drop redundant call to `or`.

src/Data/Singletons/Partition.hs
```diff
@@ -58,15 +59,17 @@ partitionDec :: DsMonad m => DDec -> m PartitionedDecs
 partitionDec (DLetDec (DPragmaD {})) = return mempty
 partitionDec (DLetDec letdec) = return $ mempty { pd_let_decs = [letdec] }
 
-partitionDec (DDataD nd _cxt name tvbs cons derivings) = do
+partitionDec (DDataD nd _cxt name tvbs k cons derivings) = do
+  extra_tvbs <- mkExtraDKindBinders k
```
goldfirere: This should use `buildDataDType`, no?

src/Data/Singletons/Partition.hs
```diff
@@ -220,3 +234,60 @@ isStockOrDefault :: Maybe DerivStrategy -> Bool
...
+--
+-- A data type is a proper GADT if one of the following conditions are met:
+--
+-- 1. A constructor has any existentially quantified type variables.
+--
+-- 2. A constructor has a context.
+--
+-- We care about this because some derivable stock classes, such as 'Enum',
+-- forbid derived instances for proper GADTs.
+isProperGADT :: forall q. DsMonad q => DType -> [DCon] -> q Bool
```
goldfirere: This is nit-picking, but this function doesn't detect proper GADTs, according to my definition of GADT. For example, it says "yes" to a type with existentials. I think GHC uses the term non-vanilla for these. I suppose it would be more evocative to call them "chocolate" types. Indeed, I might start using this terminology more! Why talk about using "fancy" types or "rich" types or "finely specified" types? Talk about using chocolate types, or even adding some chocolate sauce. Example usage: "What are you so busy doing?" "I'm refactoring my interpreter to add more chocolate sauce. Now invalid constructions will be eliminated by construction."

src/Data/Singletons/Partition.hs
RyanGlScott: I pondered this, but we end up needing intermediate parts of the `buildDataDType` computation (e.g., `all_tvbs` in `data_dec` and `ty` in `derived_decs`), so I decided to just inline the definition here to avoid some redundant work.

src/Data/Singletons/Deriving/Enum.hs
RyanGlScott: Done.

src/Data/Singletons/Partition.hs
RyanGlScott: I'm unclear as to how you're distinguishing proper GADTs, non-vanilla data types, and chocolate data types. Is there an ordering on these somehow? Can you give examples of each, and examples of one that isn't an example of the other?

goldfirere: `data Ex = forall a. MkEx a` is not vanilla, but is also not a GADT. All proper GADTs are not vanilla, but some non-GADTs are also not vanilla. (I've defined chocolate = not vanilla.)

RyanGlScott: Very well. I've changed `isProperGADT` to be named `isNonVanillaDataType`, to follow existing GHC nomenclature :)

PR #307, opened 2018-02-24
Fix #160 by introducing Data.Singletons.TypeError

### LINKED ISSUE #160: Promote `error` to custom type error

RyanGlScott:
This introduces `Data.Singletons.TypeError`, which provides a drop-in replacement for the `TypeError` API from `GHC.TypeLits`. This is based off the design in https://github.com/goldfirere/singletons/issues/160#issuecomment-338246486.

### DIFF
--- a/singletons.cabal
+++ b/singletons.cabal
@@ -55,6 +55,7 @@ library
                       template-haskell,
                       containers >= 0.5,
                       th-desugar >= 1.9 && < 1.10,
+                      pretty,
                       syb >= 0.4,
                       text >= 1.2,
                       transformers >= 0.5.2
@@ -104,6 +105,7 @@ library
                       Data.Promotion.Prelude.Show
                       Data.Promotion.Prelude.Tuple
                       Data.Promotion.Prelude.Void
+                      Data.Singletons.TypeError
                       Data.Singletons.TypeLits
                       Data.Singletons.Decide
                       Data.Singletons.ShowSing
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -35,6 +35,9 @@ Changelog for singletons project
 
 * Permit singling of expression and pattern signatures.
 
+* Add `Data.Singletons.TypeError`, which provides a drop-in replacement for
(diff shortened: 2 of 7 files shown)


singletons.cabal
```diff
@@ -55,6 +55,7 @@ library
                       template-haskell,
                       containers >= 0.5,
                       th-desugar >= 1.9 && < 1.10,
+                      pretty,
```
RyanGlScott: Note that this is shipped with GHC, so this doesn't actually add a new dependency _per se_.

LeanderK:
The ticket, that this pull request is closing, is named "Promote `error` to custom type error", but this pull-request is not promoting `error` to a custom type error (?). I have read the discussion in #160, but I am not sure whether we are not promoting `error` to a custom type error because of technical limitations or because it's a deliberate design decision.

RyanGlScott:
It's a deliberate design decision. `TypeError` is quite buggy in many ways (it often kicks in [too early](https://ghc.haskell.org/trac/ghc/ticket/14771) or [too late](https://ghc.haskell.org/trac/ghc/ticket/13775)), and I'd prefer to make it opt-in.

Moreover, universally promoting `error` to `TypeError` would be problematic since we'd have to export `TypeError` from the `singletons` Preludes, but as noted before, our version of `TypeError` clashes with the `TypeError` from `GHC.TypeLits`, which would likely result in mass confusion. I'd rather let users decide whether they want to do this or not.

LeanderK:
Got it, thanks!

PR #314, opened 2018-03-04
Fix #184 by adding support for monads 

### LINKED ISSUE #184: Support monads

RyanGlScott:
At long last, this adds supports for `Monad` and friends. This means we can properly promote and single:

* `do`-notation
* List comprehensions
* `ParallelListComp` (through the `Data.Singletons.Prelude.Monad.Zip` module)

Nothing needed to be changed from the code generation side to make this work—all that was needed was some minor tweaks to the original source code:

* Some extra kind signatures had to be inserted to avoid CUSK issues (e.g., `class Functor (f :: Type -> Type)`).
* Some definitions which rely on infinite lists, such as `some`, `many`, and `forever`, cannot be promoted.

One thing this patch does _not_ do (since it's already pretty large) is add support for `Foldable` and `Traversable`. As a result, I've left some exports from `Data.Singletons.Prelude.Monad` commented out, since they're supposed to be re-exported from `Traversable`.

### DIFF
--- a/src/Data/Singletons/Prelude/Monad/Internal.hs
+++ b/src/Data/Singletons/Prelude/Monad/Internal.hs
@@ -1,0 +1,511 @@
+{-# LANGUAGE DefaultSignatures #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE InstanceSigs #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeInType #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Data.Singletons.Prelude.Monad.Internal
+-- Copyright   :  (C) 2018 Ryan Scott
+-- License     :  BSD-style (see LICENSE)
+-- Maintainer  :  Richard Eisenberg (rae@cs.brynmawr.edu)
+-- Stability   :  experimental
+-- Portability :  non-portable
+--
+-- Defines the promoted and singled versions of:
+--
+-- * Functor
+-- * Applicative
+-- * Alternative
+-- * Monad
+-- * MonadPlus
+--
+-- As well as auxiliary definitions.
+--
+-- This module exists to break up import cycles.
+--
+----------------------------------------------------------------------------
+
+module Data.Singletons.Prelude.Monad.Internal where
+
+import Data.Kind
+import Data.List.NonEmpty (NonEmpty(..))
+import Data.Singletons.Prelude.Base
+import Data.Singletons.Prelude.Instances
+import Data.Singletons.Single
+import Data.Singletons.TypeLits.Internal
+
+$(singletonsOnly [d|
+  infixl 4  <$
+
+  {- -| The 'Functor' class is used for types that can be mapped over.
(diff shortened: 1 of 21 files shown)


RyanGlScott:
Ping?

src/Data/Singletons/Prelude/Monad/Internal.hs
```diff
@@ -0,0 +1,511 @@
...
+      --
+      -- Some functors support an implementation of 'liftA2' that is more
+      -- efficient than the default one. In particular, if 'fmap' is an
+      -- expensive operation, it is likely better to use 'liftA2' than to
+      -- 'fmap' over the structure and then use '<*>'.
+      liftA2 :: (a -> b -> c) -> f a -> f b -> f c
+      liftA2 f x = (<*>) (fmap f x)
+
+      -- -| Sequence actions, discarding the value of the first argument.
+      (*>) :: (f :: Type -> Type) a -> f b -> f b
```
goldfirere: We should document why `f` needs a kind signature here.

src/Data/Singletons/Prelude/Monad/Zip.hs
```diff
@@ -0,0 +1,99 @@
...
+{-# LANGUAGE TypeInType #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Data.Singletons.Prelude.Monad.Zip
+-- Copyright   :  (C) 2018 Ryan Scott
+-- License     :  BSD-style (see LICENSE)
+-- Maintainer  :  Richard Eisenberg (rae@cs.brynmawr.edu)
```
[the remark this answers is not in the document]
RyanGlScott: I'll do this separately in https://github.com/goldfirere/singletons/issues/331.

src/Data/Singletons/Prelude/Monad/Internal.hs
RyanGlScott: I've added a lengthy `Note [How to get the right kinds when promoting Functor and friends]` to explain the subtleties involved.

mstksg:
Just realized that this also opens the door for declaring Van Laarhoven Lenses :)

PR #344, opened 2018-06-21
Support deriving Functor, Foldable, and Traversable

### LINKED ISSUE #184: Support monads

RyanGlScott:
This implements the last bullet point in #184: supporting `DeriveFunctor`, `DeriveFoldable`, and `DeriveTraversable` in `singletons`. Much of the code in this commit is cargo-culted from the GHC source code, since the algorithms that it employs to derive those classes work much the same in `singletons`.

Fixes #184 for good.


src/Data/Singletons/Partition.hs
```diff
@@ -224,6 +235,18 @@ partitionDeriving mb_strat deriv_pred mb_ctxt ty data_decl =
        , deriv_name == enumName
       -> mk_derived_inst <$> mk_instance mkEnumInstance
 
+       | stock_or_default
```
goldfirere: This region could be cleaned up a bit by having an mapping (or an associative list, at list) from class names to instance makers.

src/Data/Singletons/Prelude/Functor.hs
```diff
@@ -194,13 +195,8 @@ $(singletonsOnly [d|
...
   void x = () <$ x
 
-  -- deriving instance Functor ((,) a)
-  instance Functor ((,) a) where
-      fmap f (x,y) = (x, f y)
-
-  -- deriving instance Functor Down
-  instance Functor Down where
-    fmap f (Down x) = Down (f x)
+  deriving instance Functor ((,) a)
```
goldfirere: Why are these here? Maybe there are module dependencies around `Down`, but the tuple instance should be in `Monad.Internal`, no?

src/Data/Singletons/Prelude/Identity.hs
```diff
@@ -81,9 +83,7 @@ $(singletonsOnly [d|
     showsPrec d (Identity x) = showParen (d > 10) $
       showString "Identity " . showsPrec 11 x
 
-  -- deriving instance Functor Identity
-  instance Functor Identity where
-    fmap f (Identity x) = Identity (f x)
+  deriving instance Functor Identity
```
goldfirere: `Foldable`? `Traversable`?

src/Data/Singletons/Prelude/Monoid.hs
```diff
@@ -178,9 +179,7 @@ $(singletonsOnly [d|
     pure = First . pure
     First f <*> First x = First (f <*> x)
 
-  -- deriving instance Functor First
-  instance Functor First where
-    fmap f (First x) = First (fmap f x)
+  deriving instance Functor First
```
goldfirere: `Traversable`?

src/Data/Singletons/Prelude/Monoid.hs
```diff
@@ -198,9 +197,7 @@ $(singletonsOnly [d|
     pure = Last . pure
     Last f <*> Last x = Last (f <*> x)
 
-  -- deriving instance Functor Last
-  instance Functor Last where
-    fmap f (Last x) = Last (fmap f x)
+  deriving instance Functor Last
```
goldfirere: `Traversable`?

src/Data/Singletons/Prelude/Semigroup.hs
```diff
@@ -227,9 +206,7 @@ $(singletonsOnly [d|
     enumFromTo (First a) (First b) = First `map` enumFromTo a b
     enumFromThenTo (First a) (First b) (First c) = First `map` enumFromThenTo a b c
 
-  -- deriving instance Functor First
-  instance Functor First where
-    fmap f (First x) = First (f x)
+  deriving instance Functor First
```
goldfirere: This isn't the first time I've seen this declaration...

src/Data/Singletons/Prelude/Semigroup/Internal.hs
```diff
@@ -178,9 +177,7 @@ $(singletonsOnly [d|
     pure = Sum
     Sum f <*> Sum x = Sum (f x)
 
-  -- deriving instance Functor Sum
-  instance Functor Sum where
-    fmap f (Sum x) = Sum (f x)
+  deriving instance Functor Sum
```
goldfirere: `Foldable`? `Traversable`?

src/Data/Singletons/Partition.hs
RyanGlScott: Good idea, I'll do that.

src/Data/Singletons/Prelude/Functor.hs
RyanGlScott: I suppose so. I'll try moving the `(,)` instances to `Monad.Internal` and see if anything blows up.

src/Data/Singletons/Prelude/Identity.hs
RyanGlScott: There's a good reason for this one—`Data.Foldable` and `Data.Traversable` must import `Identity`, `Const`, `Sum`, `Product`, `First`, `Last`, and several other newtypes to implement some of its functionality. (For similar reasons, `base` also employs a similar module structure.)

src/Data/Singletons/Prelude/Monoid.hs
RyanGlScott: See my comment above.

src/Data/Singletons/Prelude/Monoid.hs
RyanGlScott: See my comment above.

src/Data/Singletons/Prelude/Semigroup.hs
RyanGlScott: Note that the [`First` in `Data.Semigroup`](https://hackage.haskell.org/package/base-4.11.1.0/docs/Data-Semigroup.html#t:First) is not the same [`First` as in `Data.Monoid`](https://hackage.haskell.org/package/base-4.11.1.0/docs/Data-Monoid.html#t:First).

src/Data/Singletons/Prelude/Semigroup/Internal.hs
RyanGlScott: See my comment above.

goldfirere:
OK. I didn't do a close analysis. If you're happy with the instances that are declared, then so am I.

src/Data/Singletons/Prelude/Functor.hs
RyanGlScott: Ugh, _now_ I remember why I didn't do this before. It's because these instances use `Monoid`, but `Data.Singletons.Prelude.Monad.Internal` occurs much lower in the dependency graph than `Data.Singletons.Prelude.Monoid`. It would be possible to do even more internal module reorganization to allow this, but I'm not sure it's worth it.

goldfirere:
OK. Don't let me make busywork for you. Do what you think is best.

Issue #324, opened 2018-04-22
GHC 8.6 checklist


RyanGlScott:
This issue serves as a reminder to do certain things before we ship a new major release that supports GHC 8.6:

* [x] In 6eb213390729706e1840eaa67124caf26ac839f3 and ddd353afa3c92a53c00f22919611f734eaa0f29f, I added some CPP to temporarily support both GHC 8.4 and HEAD. We should remove this CPP once we drop support for GHC 8.4.

In addition, there are some other things that we could do if some new language extensions make it in to GHC 8.6:

* Template Haskell changes. In particular:
  * [x] Update `th-desugar` to support the new `ViaStrategy` AST form, as well as `QuantifiedConstraints`. (goldfirere/th-desugar#86)
  * [x] Just as `singletons` does not support `GeneralizedNewtypeDeriving`, it will also not support `DerivingVia`. Implement this.

* [x] If `QuantifiedConstraints` lands in 8.6, then we can completely do away with the `ShowSing` class. We can take code like this:

   ```haskell
   instance (ShowSing a, ShowSing b) => ShowSing (Either a b) where showsSingPrec = ...
   instance (ShowSing a, ShowSing b) => Show (Sing (e :: Either a b)) where showsPrec = showsSingPrec
   ```

   And replace it with a much more concise equivalent:

   ```haskell
   type ShowSing k = (forall (z :: k). Show (Sing z) :: Constraint)
   deriving instance (ShowSing a, ShowSing b) => Show (Sing (e :: Either a b))
   ```

   See the `quantified-constraints` branch for a WIP implementation of this.
* [x] Now that [GHC is smarter about kind inference in CUSKs](http://git.haskell.org/ghc.git/commit/12794287174146f982257cdeffd491e3e23838dc), we may be able to remove some superfluous kind signatures on `k -> Type`-kinded things (e.g., in `Functor` or `Const`).
* [x] Mention in the `README` that you'll likely need to enable `NoStarIsType` in order to profitably use `PNum(type (*))`.

RyanGlScott:
#351 and #352 implement these.

RyanGlScott:
At this point, GHC 8.6 is due to be released Real Soon Now, and I don't expect any further changes that would affect the expected test results (74176fa8e8392f21e116dd2b4b259aed2076d7c9 should be the last of them, I think). Are we ready to release `th-desugar-1.9` and `singletons-2.5` to Hackage?

goldfirere:
[...]

PR #351, opened 2018-07-10
Require GHC 8.6

RyanGlScott:
This checks off all boxes in #324 except the third one (involving `ShowSing`, which I'll tackle separately). Specifically, this:

* Removes some CPP needed for GHC 8.4.
* Bumps the `th-desugar` submodule.
* Singling `DerivingVia` does nothing (but emits a warning).
* Promoting or singling a quantified constraint results in an error.
* Now that GHC is smarter about kind inference in CUSKs, we can remove some superfluous kind signatures on certain class methods in `Applicative`, `Monad`, etc., as well as on definitions involving `Const`.
* Puts a mention in the `README` of how `NoStarIsType` is needed to use `PNum(type (*))`.


README.md
```diff
@@ -49,6 +49,7 @@ following:
 * `GADTs`
 * `InstanceSigs`
 * `KindSignatures`
+* `NoStarIsType`
```
goldfirere: And `QuantifiedConstraints`? Or is that your separate `ShowSing` patch?

RyanGlScott: Indeed, I put that in #352.

PR #352, opened 2018-07-10
Use QuantifiedConstraints to make ShowSing less ad hoc

RyanGlScott:
This implements a nice simplification permitted by the existence of `QuantifiedConstraints`. We can take the `ShowSing` class, which is wildly ad hoc, and convert it into a much more svelte quantified constraint synonym. In particular, we can take code like this:

```haskell
instance (ShowSing a, ShowSing b) => ShowSing (Either a b) where showsSingPrec = ...
instance (ShowSing a, ShowSing b) => Show (Sing (e :: Either a b)) where showsPrec = showsSingPrec
```

And replace it with a much more concise equivalent:

```haskell
type ShowSing k = (forall (z :: k). Show (Sing z) :: Constraint)
deriving instance (ShowSing a, ShowSing b) => Show (Sing (e :: Either a b))
```

This means that anyone who singles derived `Show` instances will now need to enable `QuantifiedConstraints` and `StandaloneDeriving`, but this shouldn't be too burdensome.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -31,6 +31,22 @@ Changelog for singletons project
   Haskell-generated code may require `GADTs` in situations that didn't require
   it before.
 
+* Overhaul the way derived `Show` instances for singleton types works. Before,
+  there was an awkward `ShowSing` class (which was essentially a cargo-culted
+  version of `Show` specialized for `Sing`) that one had to create instances
+  for separately. Now that GHC has `QuantifiedConstraints`, we can scrap this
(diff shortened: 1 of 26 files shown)


RyanGlScott:
> If you're reasonably confident in this, then so am I.

Sounds good!

I'll wait to land this until at least GHC 8.6.1-alpha2 is released, since some of the test suite output is likely to change due to some GHC pretty-printer changes that have yet to be merged.

Issue #356, opened 2018-08-06
GHC 8.8 checklist


RyanGlScott:
This issue serves as a reminder to do certain things before we ship a new major release that supports GHC 8.8:

* [x] Bump the `th-desugar` submodule to incorporate changes from goldfirere/th-desugar#105.
* [x] `singletons` is completely borked at the moment on GHC HEAD due to #357. Figure out how to work around this.
* [X] Adapt to the `MonadFail` proposal.
  * [ ] When we require GHC 8.8 or later to build, we should remove qualified imports to `Control.Monad.Fail`, as they will no longer be necessary due to `MonadFail` being exported by the `Prelude`.
* [X] Now that [`!`](https://gitlab.haskell.org/ghc/ghc/commit/bd7898537768f936d05c0c83eef1cd9b00933347) and [`.`](https://gitlab.haskell.org/ghc/ghc/commit/887454d8889ca5dbba70425de41d97939cb9ac60) can be parsed as type operators, we should remove the special-casing for `(!)` and `(.)`. (See [`Note [Special cases for (.) and (!)]`](https://github.com/goldfirere/singletons/blob/a9db6ff634d00a11a74595187e4ed935715f6626/src/Data/Singletons/Names.hs#L355-L361).)
* [ ] We can now rewrite `ApplyTyCon`:

   https://github.com/goldfirere/singletons/blob/ea63077a16b27f56ad5ef53461abdb57bfefc633/src/Data/Singletons/Internal.hs#L204-L206

   To use visible kind application:

   ```haskell
   type family ApplyTyCon :: (k1 -> k2) -> (k1 ~> unmatchable_fun) where
     ApplyTyCon @k1 @(k2 -> k3) @unmatchable_fun = ApplyTyConAux2
     ApplyTyCon @k1 @k2         @k2              = ApplyTyConAux1
   ```

   This change isn't strictly necessary, but it does make the right-hand sides a little less noisy.

RyanGlScott:
The third bullet point (about the `(!)` and `(.)` type operators) won't be possible unless we can devise a fix for [Trac #16339](https://ghc.haskell.org/trac/ghc/ticket/16339), which prevents us from replacing `type (:.)` with `type (.)` in `singletons`' export lists.

int-index:
The fix for #16339 has been merged to the `ghc-8.8` branch.

RyanGlScott:
[...]

PR #379, opened 2019-01-14
Bump th-desugar submodule to 9b9a6f0bdb6becd3f5ae896ddb539069131f7ba0

RyanGlScott:
This modifies `singletons` to incorporate changes from goldfirere/th-desugar#105, which modifies the `th-desugar` AST in some rather non-trivial ways:

* I had to plumb type family `Name`s through in certain places (e.g., `promoteClause`) now that `TySynEqn` requires them to be present on the LHS.
* I had to change `[DType]` to `[DTypeArg]` in a few places (e.g., `singPredRec`) since `applyDType` now requires that as an argument. In addition, I needed to introduce a `promoteTypeArg` function to use in `singPredRec`. On the other hand, I was able to remove `singletons`'s `unfoldType` function.

Despite the fact that the `th-desugar` commit is titled "GHC 8.8 support", this does _not_ change `singletons` to support GHC 8.8 yet. This is simply a preliminary patch en route to that.

Addresses one bullet point of #356.


goldfirere:
I'm afraid I've spent my time budget for singletons/th-desugar for some time. If there's something specific that wants my attention, please email me. Otherwise, I trust your judgment. Thanks!

PR #389, opened 2019-04-24
Use fromIntegral to avoid GHCJS conflict

matthewbauer:
This fixes an issue I am getting on GHCJS:

```

src/Data/Singletons/Util.hs:69:16: error:
    • Couldn't match type ‘Integer’ with ‘Int’
      Expected type: q Int
        Actual type: q Uniq
    • In the expression: return n
      In a case alternative: NameU n -> return n
      In a stmt of a 'do' block:
        case flav of
          NameU n -> return n
          _ -> error "Internal error: `qNewName` didn't return a NameU"
   |
69 |     NameU n -> return n
   |                ^^^^^^^^
```

### DIFF
--- a/src/Data/Singletons/Util.hs
+++ b/src/Data/Singletons/Util.hs
@@ -96,7 +96,7 @@ qNewUnique :: DsMonad q => q Int
 qNewUnique = do
   Name _ flav <- qNewName "x"
   case flav of
-    NameU n -> return n
+    NameU n -> return $ fromIntegral n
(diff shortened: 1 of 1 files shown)


RyanGlScott:
Thanks for the patch! I had no idea that GHCJS shipped a different version of the `template-haskell` API than GHC did.

I'll go ahead and merge this. Some additional questions:

* Did you experience any other issues with `singletons` on GHCJS (e.g., did the test suite pass)?
* Do you need a new release on Hackage with this fix?

PR #390, opened 2019-04-27
 Remove special cases for (.) and (!)

int-index:
This change is split into two commits:

1. Adapt the test suite to GHC 8.8, now that an [alpha](https://www.haskell.org/ghc/blog/20190425-ghc-8.8.1-alpha1-released.html) has been released.
2. Use the dot/bang type operators that GHC 8.8 learned to parse.


int-index:
The CI failure is due to version bounds. I worked around it locally by using `--allow-newer`, not sure what to do in `.travis.yml` instead, it is quite convoluted.

.travis.yml
```diff
@@ -30,8 +30,8 @@ before_cache:
 
 matrix:
   include:
-    - compiler: "ghc-8.6.4"
-      addons: {apt: {packages: [ghc-ppa-tools,cabal-install-2.4,ghc-8.6.4], sources: [hvr-ghc]}}
+    - compiler: "ghc-8.8.1"
+      addons: {apt: {packages: [ghc-ppa-tools,cabal-install-2.4,ghc-8.8.1], sources: [hvr-ghc]}}
```
RyanGlScott: Don't edit the `.travis.yml` file directly. Instead, let's autogenerate this with [`haskell-ci`](https://github.com/haskell-CI/haskell-ci). It should suffice to do the following:

1. Build `haskell-ci` from source at [its GitHub repo](https://github.com/haskell-CI/haskell-ci)
2. Change the `tested-with` stanza in `singletons.cabal` to be `8.8.1`
3. Run `haskell-ci regenerate` in the same directory as `singletons.cabal`

This will also fix the version bounds problem you observed in https://github.com/goldfirere/singletons/pull/390#issuecomment-487273812.

tests/compile-and-dump/Promote/GenDefunSymbols.ghc88.template
```diff
@@ -9,7 +9,7 @@ Promote/GenDefunSymbols.hs:0:0:: Splicing declarations
       where
         LiftMaybeSym1KindInference :: forall f0123456789876543210
                                              x0123456789876543210
-                                             arg. Data.Singletons.Internal.SameKind (Apply (LiftMaybeSym1 f0123456789876543210) arg) (LiftMaybeSym2 f0123456789876543210 arg) =>
+                                             arg. singletons-2.6:Data.Singletons.Internal.SameKind (Apply (LiftMaybeSym1 f0123456789876543210) arg) (LiftMaybeSym2 f0123456789876543210 arg) =>
```
RyanGlScott: I doubt that this is directly the result of your changes, but it's a bit worrying that the `singletons-2.6` tag is creeping into test output like this, since it means that we'll have to update golden files every time the version number changes. I thought that [this code](https://github.com/goldfirere/singletons/blob/ea5787091689e857ea7c71a89ac2823644e493bd/tests/SingletonsTestSuiteUtils.hs#L175-L178) was stripping out the version numbers—do you have any idea why it isn't doing so now?

tests/compile-and-dump/Singletons/FunctorLikeDeriving.ghc88.template
```diff
@@ -169,16 +169,16 @@ Singletons/FunctorLikeDeriving.hs:(0,0)-(0,0): Splicing declarations
...
       Fmap_0123456789876543210 _f_0123456789876543210 (MkT2 a_0123456789876543210) = Apply MkT2Sym0 (Apply (Apply (Apply Lambda_0123456789876543210Sym0 _f_0123456789876543210) a_0123456789876543210) a_0123456789876543210)
-    type Fmap_0123456789876543210Sym2 (a0123456789876543210 :: (~>) a0123456789876543210 b0123456789876543210) (a0123456789876543210 :: T x0123456789876543210 a0123456789876543210) =
+    type Fmap_0123456789876543210Sym2 (a0123456789876543210 :: (~>) a01234567898765432109 b01234567898765432109) (a0123456789876543210 :: T x0123456789876543210 a01234567898765432109) =
```
RyanGlScott: Again, this isn't directly your fault, but it's rather curious that these unique numbers became larger all of a sudden. I wonder if there's some underlying explanation for why this is happening? (If you don't have an answer for this question, don't worry about it too much.)

tests/compile-and-dump/Promote/GenDefunSymbols.ghc88.template
int-index: > do you have any idea why it isn't doing so now?

I don't know if I'm running the test suite correctly, the command I use is `cabal v2-test --allow-newer`. Could that be the problem?

RyanGlScott: That's exactly how I run the test suite as well, so I doubt that that's the problem.

tests/compile-and-dump/Singletons/FunctorLikeDeriving.ghc88.template
int-index: This happened just from using GHC 8.8, but I haven't bisected what GHC change caused it.

.travis.yml
int-index: Thanks, I'll do that.

tests/compile-and-dump/Singletons/FunctorLikeDeriving.ghc88.template
RyanGlScott: One possibility is that something related to GHC's generation of unique names changed between 8.6 and 8.8. A similar increase in unique number size happened between 8.0 and 8.2 when [this commit] landed (see [here](https://github.com/goldfirere/singletons/commit/637fb4d0435220f740bc658973f7a05a499dcee6#diff-7b178c8c1e207617509cdfd8802845faR4) for an example).

int-index:
> Which approach would you prefer? 

Yes, I've noticed the existence of `ghc-8.8`, only a bit too late. Still, I usually prefer to do one thing at a time, so let's land the type operator changes separately.

singletons.cabal
```diff
@@ -16,10 +16,10 @@ extra-source-files: README.md, CHANGES.md,
...
-                    tests/compile-and-dump/InsertionSort/*.ghc86.template,
-                    tests/compile-and-dump/Promote/*.ghc86.template,
-                    tests/compile-and-dump/Singletons/*.ghc86.template
+                    tests/compile-and-dump/GradingClient/*.ghc88.template,
+                    tests/compile-and-dump/InsertionSort/*.ghc88.template,
+                    tests/compile-and-dump/Promote/*.ghc88.template,
+                    tests/compile-and-dump/Singletons/*.ghc88.template
 license:        BSD3
 license-file:   LICENSE
 build-type:     Custom
```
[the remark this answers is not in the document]
int-index: Done.

.travis.yml
int-index: The 8.8 job was added to `allow_failures` by the script – do I fix it up manually?

tests/compile-and-dump/Promote/GenDefunSymbols.ghc88.template
int-index: Seems to be a local misconfiguration, on CI the `singletons-2.6` tag is stripped off: https://travis-ci.org/goldfirere/singletons/jobs/525356545

RyanGlScott: Interesting! I wonder if perhaps this is caused by your machine having a different version of `sed`? As one example, we currently have to employ [some CPP](https://github.com/goldfirere/singletons/blob/ea5787091689e857ea7c71a89ac2823644e493bd/tests/SingletonsTestSuiteUtils.hs#L163-L167) to work around differences between the Linux and macOS versions of `sed`.

int-index: macOS `sed` makes me sad. I solved the issue by using `sed` from `nixpkgs`

RyanGlScott: Bah, so it _is_ a macOS issue. In that case, can you open a separate issue about this? I really hate having platform-dependent behavior affecting whether the test suite passes, so perhaps we can replace the use of `sed` with something that has more predictable behavior.

.travis.yml
RyanGlScott: I vote to keep it the way it is. A number of `singletons`' dependencies only build at the moment since they have `head.hackage` patches, but if a newer version of a dependency were to be released and `head.hackage` lagged behind the latest Hackage version, then it's possible that the build could start to break if Travis picked a build plan that didn't include the patched version.

`haskell-ci` usually switches off the `GHCHEAD` behavior of a pre-release version of GHC once it has been released in full, so we can revisit this once that occurs.

tests/compile-and-dump/Promote/GenDefunSymbols.ghc88.template
int-index: I created #391

.travis.yml
int-index: Hm, that's not the answer I expected, ha-ha. My question was more along the lines of "do I fix it up manually or via some config option". Not sure if it's a good idea to have all jobs in `allow_failures`, it will be hard to notice if a commit breaks something.

Let's keep it removed until we start seeing spurious failures, I'd rather deal with a false positive than a false negative.

RyanGlScott: Very well. If you're going to change the generated `.travis.yml`, then please add a `travis.yml.patch` file containing that change and put `travis-patches: travis.yml.patch` in `cabal.haskell-ci`. That way, this _post facto_ change will always be applied automatically instead of having to remember to do it each time the script is regenerated. (Since we have to regenerate the script every time the `th-desugar` submodule is updated, this occurs fairly regularly.)

As a reference, here is the [`travis.yml.patch` file](https://github.com/haskell-compat/base-compat/blob/c009c983308bcefb2ca026c81db8ddfd19ca4df6/travis.yml.patch) for `base-compat`.

int-index: Thanks for the suggestion, I added a `travis.yml.patch` file.

int-index:
Thank you for the review and suggestions, @RyanGlScott!

RyanGlScott:
No problem.

As far as the remaining additions from the `ghc-8.8` branch go, do you want to submit a PR for those yourself, or should I assign that task to myself?

int-index:
This patch I wrote to put my GHC parser fixes to use, and when I started I did not realize that you had a WIP branch of your own. Sure, I'll be happy to rebase the `ghc-8.8` branch if it helps, but if you intend more changes then perhaps the most efficient thing is for me not to interfere :)

RyanGlScott:
I'm perfectly happy to push the remaining changes through. I just wanted to make sure that I'd communicated that clearly so as to avoid duplicating work.

Issue #366, opened 2018-10-10
Singletons of singletons


RyanGlScott:
Can you single a singleton? Currently the answer is "no", since there's no `Sing` instances for `Sing`s. But should there be? Recently, I found myself wanting something like this when trying to write a `Sing` instance for `Sigma`. As a refresher, here's the definition of `Sigma`:

```haskell
data Sigma (s :: Type) :: (s ~> Type) -> Type where
  (:&:) :: forall s t fst. Sing (fst :: s) -> t @@ fst -> Sigma s t
```

Here's what a `Sing` instance for `Sigma` ought to look like:

```haskell
data instance Sing :: forall s (t :: s ~> Type). Sigma s t -> Type where
  (:%&:) :: forall s t (fst :: s) (sfst :: Sing fst) (snd :: t @@ fst).
            Sing sfst -> Sing snd -> Sing (sfst :&: snd :: Sigma s t)
```

However, notice that we have a field of type `Sing (sfst :: Sing fst)`—a singleton of a singleton! Thus, while you can define a `Sing` instance for `Sigma`, it's currently impossible to use in practice, since you can't reasonably construct things of type `Sing (sfst :: Sing fst)`.

I used to think that in order to make this work, I had to define instances like this:

```haskell
data instance Sing :: forall (b :: Bool). Sing b -> Bool where
  SSFalse :: Sing SFalse
  SSTrue  :: Sing STrue
```

This is unfortunate, since:

1. This would require essentially doubling the number of `Sing` instances in use today. Plus, if we ever needed a singleton of a singleton of a singleton (e.g., `SSSFalse`), then we'd need to _triple_ the number of `Sing` instances in use, and so on.
2. As of GHC 8.6, you can't actually promote data family instance constructors to the type level, so that makes this entire plan impossible.

However, there is another way forward here. You can define a single instance (no pun intended) for all `Sing`s like this:

```haskell
newtype instance Sing :: forall k (a :: k). Sing a -> Type where
  SSing :: forall k (a :: k) (s :: Sing a). Sing a -> Sing s
type SSing = (Sing :: Sing (a :: k) -> Type)
```

That's it! Now you can profitably use the `Sing` instance for `Sigma` like so:

```haskell
type family ProjSigma1 (sig :: Sigma s t) :: s where
  ProjSigma1 ((_ :: Sing fst) :&: _) = fst

projSigma2 :: forall s t (sig :: Sigma s t).
              SingKind (t @@ ProjSigma1 sig)
[...]

PR #393, opened 2019-04-29
Make Sing a type family

### LINKED ISSUE #318: Make Sing a type family
Currently `Sing` is a data family. But it might be more natural to make it a type family, for two reasons.

One reason is extremely practical. After a [recent GHC commit](http://git.haskell.org/ghc.git/commit/ffb2738f86c4e4c3f0eaacf0a95d7326fdd2e383), compiling the test files in the test suite on GHC HEAD is now agonizingly slow—to the point where Travis times out, in fact. It's all due to the fact that GHC's pattern-match coverage checker is currently ill equipped to deal with data families, and after that commit, we now _actually_ coverage-check generated definitions involving `Sing` data family instances, and that overwhelms the coverage checker. Thus, turning `Sing` into a type family would conveniently avoid that issue.

Another reason is that it might actually make the presentation of `Sing` much cleaner. In the early days of `singletons` (according to @goldfirere's recollection), `Sing` was chosen to be a data family because of the injectivity properties it had. But we could get the same injectivity properties if we just use:

```haskell
type family Sing :: k -> Type
```

Then, instead of generating something like this for each `Sing` instance (using the data family presentation):

```haskell
data instance Sing :: Bool -> Type where
  SFalse :: Sing False
  STrue :: Sing True
type SBool = (Sing :: Bool -> Type)
```

We would generate this:

```haskell
data SBool :: Bool -> Type where
  SFalse :: SBool False
  STrue :: SBool True
type instance Sing = SBool
```

Moreover, with a type family version of `Sing`, one can get away with this:

```haskell
type instance Sing = (TypeRep :: TYPE rep -> Type)
```

Which is pretty slick.

RyanGlScott:
This fixes #318 by switching `Sing` from a data family to a type family. There are several knock-on changes as a result of this switch—refer to `CHANGES.md` for an overview.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -40,14 +40,21 @@ Changelog for singletons project
     This is unlikely to affect you unless you define 'Show' instances for
     singleton types by hand. For more information, refer to the Haddocks for
     `ShowSing'` in `Data.Singletons.ShowSing`.
-  * It is no longer possible to define a single `TestEquality` and
-    `TestCoercion` instance that covers all `Sing`s, as one cannot put type
-    families inside of instance heads. Instead, `singletons` now generates a
-    separate `TestEquality`/`TestCoercion` instance for every data type that
-    singletons a derived `Eq` instance. In addition, the
-    `Data.Singletons.Decide` module now provides top-level
-    `decideEquality`/`decideCoercion` functions which provide the behavior
-    of `testEquality`/`testCoercion`, but monomorphized to `Sing`.
+  * GHC does not permit type class instances to mention type families, which
+    means that it is no longer possible to define instances that mention the
+    `Sing` type constructor. For this reason, a `WrappedSing` data type (which
+    is a newtype around `Sing`) was introduced so that one can hang instances
+    off of it.
+
+    This had one noticeable effect in `singletons`
+    itself: there are no longer `TestEquality Sing` or `TestCoercion Sing`
+    instances. Instead, `singletons` now generates a separate
+    `TestEquality`/`TestCoercion` instance for every data type that singles a
+    derived `Eq` instance. In addition, the `Data.Singletons.Decide` module
+    now provides top-level `decideEquality`/`decideCoercion` functions which
+    provide the behavior of `testEquality`/`testCoercion`, but monomorphized
+    to `Sing`. Finally, `TestEquality`/`TestCoercion` instances are provided
+    for `WrappedSing`.
 * GHC's behavior surrounding kind inference for local definitions has changed
   in 8.8, and certain code that `singletons` generates for local definitions
   may no longer typecheck as a result. While we have taken measures to mitigate
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,50 @@ Changelog for singletons project
 2.6
 ---
 * Require GHC 8.8.
+* `Sing` has switched from a data family to a type family. This has a number of
+  consequences:
+  * Names like `SBool`, `SMaybe`, etc. are no longer type synonyms for
+    particular instantiations of `Sing` but are instead the names of the
+    singleton data types themselves. In other words, previous versions of
+    `singletons` would provide this:
+
+    ```haskell
+    data instance Sing :: Bool -> Type where
+      SFalse :: Sing False
+      STrue  :: Sing True
+    type SBool = (Sing :: Bool -> Type)
+    ```
+
+    Whereas with `Sing`-as-a-type-family, `singletons` now provides this:
+
+    ```haskell
+    data SBool :: Bool -> Type where
+      SFalse :: SBool False
+      STrue  :: SBool True
+    type instance Sing @Bool = SBool
+    ```
+  * The `Sing` instance for `TYPE rep` in `Data.Singletons.TypeRepTYPE` is now
+    directly defined as `type instance Sing @(TYPE rep) = TypeRep`, without the
+    use of an intermediate newtype as before.
+  * Due to limitations in the ways that quantified constraints and type
+    families can interact
+    (see [this GHC issue](https://gitlab.haskell.org/ghc/ghc/issues/14860)),
+    the internals of `ShowSing` has to be tweaked in order to continue to
+    work with `Sing`-as-a-type-family. One notable consequence of this is
+    that `Show` instances for singleton types can no longer be derived—they
+    must be written by hand in order to work around
+    [this GHC bug](https://gitlab.haskell.org/ghc/ghc/issues/16365).
+    This is unlikely to affect you unless you define 'Show' instances for
+    singleton types by hand. For more information, refer to the Haddocks for
+    `ShowSing'` in `Data.Singletons.ShowSing`.
+  * It is no longer possible to define a single `TestEquality` and
+    `TestCoercion` instance that covers all `Sing`s, as one cannot put type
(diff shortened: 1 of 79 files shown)


int-index:
I was initially opposed to the change, but then I realized: the important part of this encoding is that it says

```
type family Sing :: k -> Type
```

rather than

```
type family Sing (a :: k) :: Type
```

This means we can still partially apply `Sing`.

Would be good to see a mention of this in the docs (or a Note)

int-index:
I wonder, would this injectivity annotation make any difference?

```
type family Sing = (r :: k -> Type) | r -> k
```

RyanGlScott:
I've just added a `Note [The kind of Sing]` underneath the definition of `Sing` which explains why we don't pick something like `type family Sing (a :: k)`.

> I wonder, would this injectivity annotation make any difference?

It wouldn't. Since `Sing` is declared to have return kind `k -> Type`, it is automatically an injective type constructor as a consequence, so an injectivity declaration would be redundant. (I explain this in `Note [The kind of Sing]` as well.)

RyanGlScott:
Since this is a nontrivial redesign of some of `singletons`' fundamentals, I'm currently doing some smoke-testing of other libraries that depend on `singletons` to see if there are other problems that I didn't anticipate. In this process, I did encounter one small wrinkle that I overlooked when I wrote this patch. Previously, `singletons` would generate this data family instance:

```hs
data instance Sing :: D a -> Type where
```

It now generated this equivalent data type:

```hs
data SD :: D a -> Type
```

At least, I had previously convinced myself that they were equivalent. But I now realize that there's actually a subtle difference between the two in that the latter type does not have a CUSK! Data family instances don't really have an equivalent of CUSKs, but since data families themselves always have CUSKs, it not too much of a stretch to say that former type _does_ have a CUSK.

That being said, given that `singletons` doesn't really support [singling GADTs](https://github.com/goldfirere/singletons/issues/150) at the moment, I can't think of any `singletons` code that you can successfully compile today where the lack of a CUSK would be an issue. On the other hand, I do have a [`singleton-gadts`](https://github.com/RyanGlScott/singleton-gadts) repo—which piggybacks off of `singletons`' Template Haskell machinery—that fails to compile with this patch. That's because `singleton-gadts` tries to generate (or rather, the underlying `singletons` machinery tries to generate) the following data type:

```hs
data (%:~:) :: a :~: b -> Type where
  SRefl :: (%:~:) Refl
```

Lacking a CUSK, `(%:~:)` fails to compile. Granted, I could take the `singletons`-generated `Dec` and manually insert the `forall a b.` into `(%:~:)`'s kind _post facto_, but it feels like it would be much more direct to do this in `singletons` itself. Unless someone feels strongly otherwise, I think I will go ahead and do so.

goldfirere:
A very quick gander at the code makes this look quite a bit easier than I would have thought. Cool. I'm in full support of this change, unsurprisingly. :)

RyanGlScott:
Hm, it turns out that @mstksg's `decidable` library doesn't compile with this patch either. Here is a boiled-down example:

```hs
{-# LANGUAGE AllowAmbiguousTypes #-}
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE PolyKinds #-}
{-# LANGUAGE RankNTypes #-}
module Foo where

import Data.Kind
import Data.Singletons
import Data.Singletons.Decide

type Predicate k = k ~> Type
type TyPred = (TyCon1 :: (k -> Type) -> Predicate k)
type Evident = (TyPred Sing :: Predicate k)
type Decide p = forall a. Sing a -> Decision (p @@ a)

class Decidable p where
    decide :: Decide p
instance Decidable Evident where
    -- decide = Proved . prove @Evident
```

With `Sing`-as-a-type-family, this fails to compile with:

```
Foo.hs:20:10: error:
    • Illegal type synonym family application ‘Sing’ in instance:
        Decidable Evident
    • In the instance declaration for ‘Decidable Evident’
   |
20 | instance Decidable Evident where
   |          ^^^^^^^^^^^^^^^^^
```

(That error message isn't great, but there's a use of `Sing` buried underneath the `Evident` type synonym.)

This worries me a bit, since I'm not exactly sure how one would port over this code. Thoughts, @mstksg?

mstksg:
i'll look into it and see.  the main issue is using a type family as an instance, right?  So `instance Decidable (TyPred1 Sing)` cannot be defined for all singletons, huh.

We may be able to get around the issue by using a newtype wrapper and writing the instance over that.  I'll try to see if I can get something that works.

RyanGlScott:
Yes, I suspect that recommending the use of a newtype wrapper around `Sing` may be the best path forward here. I was planning to introduce such a `WrappedSing` separately as part of a fix for #366 as an optional goodie for power users, but it didn't occur to me until now that `WrappedSing` might be _required_ for some use cases (such as `decidable`'s).

RyanGlScott:
@mstksg, I've just pushed a change which adds a `WrappedSing` newtype around `Sing`. With this, it may be possible to define things like `type Evident = (TyPred WrappedSing :: Predicate k)`—let me know if this is sufficient for your needs.

RyanGlScott:
Ping @mstksg. Do note that I'm essentially holding back this PR until I receive some an indication of some kind from you that this will (or won't) work out for your needs, so if you'd like to see this feature land, your feedback would be appreciated :)

mstksg:
Ah, I was going to look deeper into this last week, but things did come up.  It's a bit difficult for me to test right now because I need to set up a pipeline to test with GHC head :)  I think if a newtype wrapper or some other workaround is possible, *decidable* would be okay for it.  Overall the ergonomics of general singletons usage is a higher priority for me at this moment!  I can't promise a specific day I can get to this, so feel free to move past me if you want to move things along :)

RyanGlScott:
No worries. I'm not in a _huge_ hurry to merge this, especially since it there's still time before we even see a release candidate of GHC 8.8.1 (we're still in the alpha stage). Also, us folks at the #ghc IRC channel would be happy to help get you set up with a GHC HEAD configuration—it's much less painful than it used to be, I promise :)

RyanGlScott:
FWIW, I've pushed [a fork of `decidable`](https://github.com/RyanGlScott/decidable/commit/ceddba4d185bc9e53cf5aebcfb88d1dcfd1cd323) which shows one possible way of making it work with `WrappedSing`. (There are still some questions you would need to decide as far as which places should use `Sing` and which places should use `WrappedSing`, but each possible design is the same up to wrapping/unwrapping newtypes.)

mstksg:
Thanks for this :)  Looking over the changes, it all seems like a very reasonable interface.  The only time someone would need to deal with wrapping is `Evident`, I think.  And also with sing as a type family, we can get rid of `SIndex'` completely, right?  and just use `type Sing (Index as a) = SIndex as a`, hopefully!

RyanGlScott:
> And also with sing as a type family, we can get rid of SIndex' completely, right?

Indeed. I've pushed [a commit](https://github.com/RyanGlScott/decidable/commit/03abbb617e1069fe1440f4f69d9423383e23690e) which demonstrates one way to do this.

RyanGlScott:
Judging from recent comments (especially https://github.com/goldfirere/singletons/pull/393#issuecomment-504279219) and from the general quiescence of the discussion on this PR, it looks like we have come to a consensus that this is the right approach. In light of this, I'm going to merge this. Please speak up if you feel otherwise!

PR #397, opened 2019-05-09
Promote instance contexts

### LINKED ISSUE #396: Consider promoting type class instance contexts
If you try to promote a type class instance, as in the following example:

```hs
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE GADTs #-}
{-# LANGUAGE PolyKinds #-}
{-# LANGUAGE TypeFamilies #-}
{-# OPTIONS_GHC -ddump-splices #-}
module Foo where

import Data.Functor.Identity
import Data.Singletons.Prelude.Identity
import Data.Singletons.TH

$(promoteOnly [d|
  class C a where
    m :: a -> Bool

  instance C a => C (Identity a) where
    m (Identity x) = m x
  |])
```

You'll currently generate the following code (simplified):

```hs
class PC (a :: Type) where
  type M (arg :: a) :: Bool

instance PC (Identity a) where
  type M ('Identity x) = M x
```

There's nothing explicitly _wrong_ about this code, but it is a bit strange. Although the original `C (Identity a)` instance declaration has a `C a` constraint, the promoted `PC (Identity a)` instance has no constraints at all. This works since promoted classes can only ever have associated type families, which currently never make use of any class constraints at all.

I say "currently" since the status quo here could change in the future. If a proposal to add constrained type families (e.g., [this one](https://github.com/ghc-proposals/ghc-proposals/pull/177)) is ever implemented, then the `PC (Identity a)` instance above will no longer work without a `PC a` constraint. We might be a ways away from ever implementing such a proposal, but promoting `C a =>` to `PC a =>` would make `singletons` future-compatible with it.

There is one possible concern in that `-Wno-redundant-constraints` would produce a warning that `PC a` is redundant in `instance PC a => PC (Identity a)`. But this is really a moot concern, as there are several places in `singletons` where it generates code that trips up `-Wno-redundant-constraints` (e.g., singling derived instances). Given that we already warn users not to enable `-Wno-redundant-constraints` in the `singletons` README, I wouldn't lose sleep over this.

RyanGlScott:
Fixes #396.

### DIFF
--- a/src/Data/Singletons/Promote.hs
+++ b/src/Data/Singletons/Promote.hs
@@ -152,7 +152,7 @@ promoteEqInstance name = do
   let data_ty = foldTypeTvbs (DConT name) tvbs'
   cons' <- concatMapM (dsCon tvbs' data_ty) cons
   kind <- promoteType (foldTypeTvbs (DConT name) tvbs')
-  inst_decs <- mkEqTypeInstance kind cons'
+  inst_decs <- mkEqTypeInstance Nothing kind cons'
   return $ decsToTH inst_decs
 
 promoteInstance :: DsMonad q => DerivDesc q -> String -> Name -> q [Dec]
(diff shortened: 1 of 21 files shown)


RyanGlScott:
Abandoning this PR in favor of commit 0320f99544586ef030350ef0b4a994fe7b2d29f7, which opts to just remove `promote_superclass_pred` entirely.

PR #398, opened 2019-05-29
Really fix ghcjs th conflict

matthewbauer:
My previous fix in https://github.com/goldfirere/singletons/pull/389
was not complete. Compiling for 32-bit on 64-bit machines means that
we get truncated when the template identifier exceeds 2,147,483,647.
This leads to hitting an error when actually using singletons (as
opposed to just building):

- https://github.com/ghcjs/ghcjs/issues/617

To fix, we need to use Uniq to preserve identifier uniqueness in
template haskell. This allows GHCJS to use its own Uniq type here:

- https://github.com/ghcjs/ghc/commit/f1a393ae03d7310159d27e146197d1fd99242aac

This should have the side effect of also fixing singletons for ghc
8.8.1 now that this has been merged:

- https://gitlab.haskell.org/ghc/ghc/merge_requests/948

### DIFF
--- a/src/Data/Singletons/Util.hs
+++ b/src/Data/Singletons/Util.hs
@@ -91,11 +91,11 @@ qReportError :: Quasi q => String -> q ()
 qReportError = qReport True
 
 -- | Generate a new Unique
-qNewUnique :: DsMonad q => q Int
+qNewUnique :: DsMonad q => q Uniq
 qNewUnique = do
   Name _ flav <- qNewName "x"
   case flav of
-    NameU n -> return $ fromIntegral n
+    NameU n -> return n
     _       -> error "Internal error: `qNewName` didn't return a NameU"
 
 checkForRep :: Quasi q => [Name] -> q ()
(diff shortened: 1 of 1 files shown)


matthewbauer:
Related to this ticket: https://gitlab.haskell.org/ghc/ghc/issues/14574

RyanGlScott:
Is there a reason that you closed this, @matthewbauer?

matthewbauer:
Yeah it didn't actually fix the issue. Still trying to figure out what is going on though. It's basically the same thing reported in https://github.com/ghcjs/ghcjs/issues/617. will try -ddump-splices to debug.

matthewbauer:
@RyanGlScott It looks like this is actually an issue in th-desugar. I've opened an issue here: https://github.com/goldfirere/th-desugar/issues/119

RyanGlScott:
While this PR may not have fixed the GHCJS issue you were experiencing, I like this patch solely on the basis that it avoids the possibility of truncating `Integer`s. I've applied this separately as d78e5ce3fb865ec1c1659817b0bba1a1bccc692c.

matthewbauer:
Yeah sounds good!

Issue #378, opened 2019-01-14
Figure out how to promote visible type application sensibly


RyanGlScott:
There's currently this line in `Data.Singletons.Promote`:

https://github.com/goldfirere/singletons/blob/fb5e0051591f4e30109713e237bc2f0957406a3b/src/Data/Singletons/Promote.hs#L762-L765

Darn, if only we had visible kind applications! But hey, we're getting just that in GHC 8.8! Surely this means that we can rip out this antiquated code and treat visible type applications as a first-class citizen in `singletons`...

...but not so fast. It turns out that having visible kind applications isn't enough—we also need to figure out how to use it properly. Here's a function to help illustrate some of the challenges that VTA presents:

```haskell
constBA :: forall b a. a -> b -> a
constBA x _ = x
```

`constBA` is exactly the same as the familiar `const` function, except I've deliberately swapped the conventional order that the type variables are quantified in. One interesting observation, right off the bat, is that when `constBA` is promoted to a type family:

```haskell
type family ConstBA (x :: a) (y :: b) :: a where ...
```

The kind of `ConstBA` is now `forall a b. a -> b -> a`, not `forall b a. a -> b -> a`! What's worse, there is no reliable way of swapping the order that `a` and `b` are quantified in until we get [top-level kind signatures](https://github.com/ghc-proposals/ghc-proposals/pull/54). Blast.

But that's far from the only sticky thing about this situation. Let's suppose you want to _use_ `constBA`, like in the following function:

```haskell
blah :: forall b a. [a] -> [b -> a]
blah x = map (constBA @b) x
```

How should this promote? This is what currently happens in today's `singletons`, which simply drops VTAs:

```haskell
type family Blah (x :: [a]) :: [b ~> a] where
  Blah x = Apply (Apply MapSym0 ConstBASym0) x
```

Notice that `singletons` always uses defunctionalization symbols when promoting function application, so if we want to promote `constBA @b`, this suggests that the output should be something like `ConstBASym0 @b`. But that presents its own set of challenges, because this is the code that gets generated for `ConstBASym0`:

```haskell
data ConstBASym0 :: forall a b. a ~> b ~> a
```

This happens because during defunctionalization, we take the type `a ~> b ~> a` and quantify over it by simply gathering the free variables in left-to-right order. Unfortunately, this means that we get `forall a b. <...>` instead of `forall b a. <...>`. Double blast.

[...]

PR #406, opened 2019-09-02
Properly reject higher-rank types during promotion/singling

### LINKED ISSUE #401: Figure out a policy for rank-n types
What do you think the following code does?

```hs
$(singletons [d|
  f :: (forall a. a -> a) -> b -> b
  f _ x = x
  |])
```

To my complete surprise, this code actually _compiles_. I was baffled as to how this could possibly work until I looked at the `-ddump-splices` output, which I've produced below (after some minor cleanup):

```hs
    type family F (f :: a ~> a) (x :: b) :: b where
      F _ x = x
    sF ::
      forall b a (f :: a ~> a) (x :: b).
      Sing f -> Sing x -> Sing (F f x)
    sF _ sX = sX
```

Yikes. Unlike the type of `f`, which uses higher-rank polymorphism, neither the kind of `F` nor the type of `sF` are higher-rank! It seems that `singletons` drops the nested `forall` and, by sheer accident, this allows the generated code to compile.

This is clearly less than ideal, and we should aspire to do better. But what is the right thing to do when `singletons` is presented with rank-_n_ types? I contemplated how `singletons` might be changed so that the code it generates is also rank-_n_, but there are a number of obstacles that stand in the way of this vision:

## 1(a): Defunctionalization symbols (current `singletons`)

If we were to promote the `f` example above, here is a defunctionalization symbol that would need to be generated:

```hs
data FooSym0 :: (forall a. a ~> a) ~> b ~> b
```

But note that `(forall a. a ~> a) ~> b ~> b` is the same thing as `TyFun (forall a. a ~> a) (b ~> b) -> Type`, which is decidedly impredicative. I doubt that we want to require users to enable `ImpredicativeTypes`.

## 1(b): Defunctionalization symbols (alternative encoding of `(~>)`)

In an attempt to work around 1(a), I was reminded of [this alternative way](https://github.com/goldfirere/singletons/issues/320#issue-311443055) to encode `(~>)`:

```hs
type a ~> b = a -> b -> Type
```

I thought this might be worth considering because with this encoding, `(forall a. a ~> a) ~> b ~> b` would become `(forall a. a -> a -> Type) -> (b -> b -> Type) -> Type`, which isn't impredicative anymore! Alas, this doesn't get you very far. If you try to define an instance of `Apply` using this version of `(~>)`:

```hs
type instance forall (x :: forall a. a ~> a).
  Apply FSym0 x = FSym1 x
```

Then GHC is not happy:

```
Foo.hs:33:9: error:
    • Expected kind ‘(a0 ~> a0) ~> (b0 ~> b0)’,
        but ‘FSym0’ has kind ‘(forall a. a ~> a) -> (b0 ~> b0) -> *’
    • In the first argument of ‘Apply’, namely ‘FSym0’
      In the type instance declaration for ‘Apply’
   |
33 |   Apply FSym0 x = FSym1 x
   |         ^^^^^
```

It seems that GHC is reluctant to infer a higher-rank kind for `FSym0` in that spot. I tried to make GHC come to its senses:

```hs
type instance forall (x :: forall a. a ~> a).
  Apply @(forall a. a ~> a) FSym0 x = FSym1 x
```

But that makes GHC even less happy:

```
Foo.hs:33:3: error:
    • Illegal polymorphic type: forall a. a ~> a
    • In the type instance declaration for ‘Apply’
   |
33 |   Apply @(forall a. a ~> a) FSym0 x = FSym1 x
   |   ^^^^^
```

Bah. At this point, I gave up on the idea. (Man, I wish `(~>)` were [a part of GHC itself](https://github.com/ghc-proposals/ghc-proposals/pull/242) so that we didn't have to deal with defunctionalization-related crap...)

## 2: In singled definitions

Even if you ignore the challenges that defunctionalization presents, there's still another hurdle to clear: singling. What should the singled version of this function be?

```hs
g :: (forall a. a -> a) -> b -> b
g h x = h x
```

Here is a lousy first attempt:

```hs
type family G (h :: forall a. a ~> a) (x :: b) :: b where
  forall b (h :: forall a. a ~> a) (x :: b). G h x = h @@ x

sG :: forall b (h :: forall a. a ~> a) (x :: b).
      Sing h -> Sing x -> Sing (G h x)
sG sH sX = applySing sH sX
```

Sadly, `sG` does not typecheck:

```
Foo.hs:36:25: error:
    • Couldn't match type ‘x’ with ‘t0’
      ‘x’ is a rigid type variable bound by
        the type signature for:
          sG :: forall a1 b (h :: forall a2. a2 ~> a2) (x :: b).
                Sing h -> Sing x -> Sing (G h x)
        at Foo.hs:(34,1)-(35,38)
      Expected type: Sing t0
        Actual type: Sing x
    • In the second argument of ‘applySing’, namely ‘sX’
      In the expression: applySing sH sX
      In an equation for ‘sG’: sG sH sX = applySing sH sX
    • Relevant bindings include
        sX :: Sing x (bound at Foo.hs:36:7)
        sH :: Sing h (bound at Foo.hs:36:4)
        sG :: Sing h -> Sing x -> Sing (G h x) (bound at Foo.hs:36:1)
   |
36 | sG sH sX = applySing sH sX
   |                         ^^
```

One way to work around this is to use the following type signature for `sG` instead:

```hs
sG :: forall b (h :: forall a. a ~> a) (x :: b).
      (forall a. Sing (h @a)) -> Sing x -> Sing (G h x)
sG sH sX = applySing sH sX
```

This works, but it doesn't scale well, since this trick can only work for rank-_n_ types where the `forall`s always appear in prenex form. That is to say, if you tried to single this definition:

```hs
j :: (a -> forall b. b -> b) -> a -> c -> c
j k x y = k x y
```

Then you're hosed, since there's no way to instantiate the `b` in `a ~> forall b. b ~> b` with visible kind application alone.

-----

All of these challenges lead me to believe that we should simply make any attempt to promote or single a rank-_n_ type an error. In fact, `singletons` currently implements a half-hearted attempt to do this here:

https://github.com/goldfirere/singletons/blob/27245e24d588a63620ffb7f0276345ac36f8a6ba/src/Data/Singletons/Promote/Type.hs#L27-L29

In practice, that `(DAppT (DAppT DArrowT (DForallT (_:_) _ _)) _)` case is very rarely exercised, since all top-level function types are passed to `promoteUnraveled` first, which splits apart `DArrowT`s and then promotes the individual function argument types. As a consequence, it would catch rank-3 (or above) types, but not rank-2 types.

If we do ban rank-_n_ types, what should become of types with nested `forall`s, such as in the following example?

```hs
$(singletons [d|
  const2 :: forall a. a -> forall b. b -> a
  const2 x _ = x
  |])
```

Again, this code actually compiles today, but only because `singletons` treats it as if you had written `forall a b. a -> b -> a`. Should these be banned alongside other rank-_n_ types, or should we grandfather in these sorts of functions that do anything that is explicitly higher-rank?

RyanGlScott:
`singletons` can't handle higher-rank types. You might think that these sorts of types would naturally be rejected when promoting or singling them, but that actually turns out not to be the case! (See #401 for examples of higher-rank types that squeeze through the cracks.)

Luckily, the additional machinery that `th-desuguar` adds in goldfirere/th-desugar#125 makes it much easier to detect these. Instead of using `unravelDType` to decompose a function type (which permits all sorts of higher-rank gubbins), we use a restricted version called `unravelVanillaDType`, which throws an error if anything higher-rank is detected. See `Note [Vanilla-type validity checking during promotion]` in `D.S.Promote.Type` for a full explanation of what a "vanilla" type is.

Fixes #401.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,14 @@
 Changelog for singletons project
 ================================
 
+next
+----
+* Require GHC 8.10.
+* `singletons` now does a more much thorough job of rejecting higher-rank types
+  during promotion or singling, as they cannot support them. (Previously,
+  `singletons` would sometimes accept them, often changing rank-2 types to
+  rank-1 types incorrectly in the process.)
+
 2.6
 ---
 * Require GHC 8.8.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,9 +5,9 @@ next
 ----
 * Require GHC 8.10.
 * `singletons` now does a more much thorough job of rejecting higher-rank types
-  during promotion or singling, as they cannot support them. (Previously,
-  `singletons` would sometimes accept them, often changing rank-2 types to
-  rank-1 types incorrectly in the process.)
+  during promotion or singling, as `singletons` cannot support them.
+  (Previously, `singletons` would sometimes accept them, often changing rank-2
+  types to rank-1 types incorrectly in the process.)
 
 2.6
 ---
--- a/src/Data/Singletons/Promote/Type.hs
+++ b/src/Data/Singletons/Promote/Type.hs
@@ -7,26 +7,37 @@ This file implements promotion of types into kinds.
 -}
 
 module Data.Singletons.Promote.Type
-  ( promoteType, promoteTypeArg, promoteUnraveled
+  ( promoteType, promoteType_NC
+  , promoteTypeArg_NC, promoteUnraveled
   ) where
 
 import Language.Haskell.TH.Desugar
 import Data.Singletons.Names
+import Data.Singletons.Util
 import Language.Haskell.TH
 
--- the only monadic thing we do here is fail. This allows the function
--- to be used from the Singletons module
+-- Promote a DType to the kind level.
+--
+-- NB: the only monadic thing we do here is fail. This allows the function
+-- to be used from the Singletons module.
 promoteType :: MonadFail m => DType -> m DKind
-promoteType = go []
+promoteType ty = do
+  checkVanillaDType ty
+  promoteType_NC ty
+
+-- Promote a DType to the kind level. This is suffixed with "_NC" because
+-- we do not invoke checkVanillaDType here.
+-- See [Vanilla-type validity checking during promotion].
+promoteType_NC :: MonadFail m => DType -> m DKind
+promoteType_NC = go []
   where
     go :: MonadFail m => [DTypeArg] -> DType -> m DKind
+    go []       (DForallT _fvf _tvbs ty) = go [] ty
     -- We don't need to worry about constraints: they are used to express
     -- static guarantees at runtime. But, because we don't need to do
     -- anything special to keep static guarantees at compile time, we don't
     -- need to promote them.
-    go []       (DForallT _tvbs _cxt ty) = go [] ty
-    go []       (DAppT (DAppT DArrowT (DForallT (_:_) _ _)) _) =
-      fail "Cannot promote types of rank above 1."
+    go []       (DConstrainedT _cxt ty) = go [] ty
     go args     (DAppT t1 t2) = do
       k2 <- go [] t2
       go (DTANormal k2 : args) t1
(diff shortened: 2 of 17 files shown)


CHANGES.md
```diff
@@ -1,6 +1,14 @@
 Changelog for singletons project
 ================================
 
+next
+----
+* Require GHC 8.10.
+* `singletons` now does a more much thorough job of rejecting higher-rank types
+  during promotion or singling, as they cannot support them. (Previously,
```
goldfirere: I always thought that `singletons`'s preferred pronoun was "it", not "they". Regardless, using "they" and "them" in succession referring to different antecedents is confusing.

src/Data/Singletons/Promote/Type.hs
```diff
@@ -7,26 +7,37 @@ This file implements promotion of types into kinds.
...
+  promoteType_NC ty
+
+-- Promote a DType to the kind level. This is suffixed with "_NC" because
+-- we do not invoke checkVanillaDType here.
+-- See [Vanilla-type validity checking during promotion].
+promoteType_NC :: MonadFail m => DType -> m DKind
+promoteType_NC = go []
   where
     go :: MonadFail m => [DTypeArg] -> DType -> m DKind
+    go []       (DForallT _fvf _tvbs ty) = go [] ty
```
goldfirere: The comment explains why we drop the constraint. Good. But why drop the `forall`?

src/Data/Singletons/Util.hs
```diff
@@ -251,10 +252,130 @@ ravel :: [DType] -> DType -> DType
...
+-- Note [Vanilla-type validity checking during promotion] in
+-- Data.Singletons.Promote.Type for what "vanilla" means.)
+-- If a non-vanilla construct is encountered while decomposing the function
+-- type, an error is thrown.
+--
+-- This should be contrasted with the 'unravelDType' function from
+-- @th-desugar@, which supports the full gamut of function types. @singletons@
+-- only supports a subset of these types, which is why this function is used
+-- to decompose them instead.
+unravelVanillaDType :: DType -> ([DTyVarBndr], DCxt, [DType], DType)
```
goldfirere: Should this and the next function be combined? I suppose that would be less performant, but the duplication of code is annoying. Perhaps laziness means that the performance hit would be minimal.

src/Data/Singletons/Promote/Type.hs
RyanGlScott: Good question. Now that we've separated out `DForallT` and `DConstrainedT`, I don't think there's any reason to drop explicit `DForallT`s any more... at least in theory. In practice, I think this code path is unused at the moment, since if I apply this change:

```diff
diff --git a/src/Data/Singletons/Promote/Type.hs b/src/Data/Singletons/Promote/Type.hs
index d35ac75..d9f893d 100644
--- a/src/Data/Singletons/Promote/Type.hs
+++ b/src/Data/Singletons/Promote/Type.hs
@@ -32,7 +32,9 @@ promoteType_NC :: MonadFail m => DType -> m DKind
 promoteType_NC = go []
   where
     go :: MonadFail m => [DTypeArg] -> DType -> m DKind
-    go []       (DForallT _fvf _tvbs ty) = go [] ty
+    go []       (DForallT fvf tvbs ty) = do
+      ty' <- go [] ty
+      pure $ DForallT fvf tvbs ty'
     -- We don't need to worry about constraints: they are used to express
     -- static guarantees at runtime. But, because we don't need to do
     -- anything special to keep static guarantees at compile time, we don't
```

Then the test suite continues to pass without needing any changes whatsoever. As luck would have it, `promoteType` only ever seems to be called on types without top-level `forall`s. For example, `D.S.Single.Type.singType`, which singles top-level type signatures, splits the type into argument types (discarding top-level `forall`s) and then invokes `promoteType` on each argument type individually.

I could foresee a future in which preserving `DForallT` in `promoteType` is important, however. If we implement #378, for instance, we'll need to take this function:

```hs
blah :: forall b a. [a] -> [b -> a]
blah = ...
```

And promote it to this:

```hs
type Blah :: forall b a. [a] -> [b ~> a]
type family Blah (x :: [a]) :: [b ~> a] where ...
```

The standalone kind signature for `Blah` will be the output of `promoteType`, so in order to ensure that `forall b a.` appears in that SAK (and in exactly that order, to preserve its properties _vis-à-vis_ `TypeApplications`), `promoteType` needs to preserve `forall`s. For this reason, I've gone ahead and just applied the change above.

goldfirere: Sounds good. :)

src/Data/Singletons/Util.hs
goldfirere: That's why I was worried about performance. Of course, `checkVanillaDType` could just call `isJust . unravelVanillaDType`...

RyanGlScott: Actually, I think I've come around to the idea of combining these two functions, since they're just _so_ darn similar. And I don't think the performance hit would be that noticeable.

I've pushed a commit which implements this merger. Let me know if you like the way it looks—I did some refactoring along the way to make it look somewhat nicer.

goldfirere: Ah yes, much better now. Thanks!

goldfirere:
Ship it!

PR #409, opened 2019-09-05
Remove accept-all script in favor of tasty's --accept flag

RyanGlScott:
The `accept-all` script exists for one purpose: to accept changes in expected test suite output. But it turns out that `tasty-golden` already supports doing this with its `--accept` flag. This allows for integrating with your build tool itself, since you can simply say:

```
cabal test --test-options="--accept"
```

`singletons`' test suite is slightly nontrivial, so we need to perform a little bit of footwork to make `--accept` work nicely with the test suite infrastructure. Thankfully, it's not too bad.


tests/README.md
```diff
@@ -54,12 +54,20 @@ tests:
...
 
    runs all tests in the Testsuite/Singletons branch of the test tree.
    SingletonsTestSuite module defines structure of the test tree.
 
+ * If you modify `singletons`, you may cause the actual output of some tests to
+   change. If these changes are what you intended, you can accept the new
+   output by running the following command:
+
+   ```bash
+   cabal v2-test --test-options="--accept"
```
int-index: the v2 prefix is redundant with the recently released cabal 3.0

RyanGlScott: My original thought was that this command would be backwards-compatible with older `cabal-install`s, but I now realize that GHC 8.8 requires `cabal-install-3.0` or later anyway.

I've shortened this to just `cabal test`.

PR #413, opened 2019-09-09
Handle type constructors' fixity declarations more carefully

### LINKED ISSUE #412: A collage of fixity declaration bugs
If you compile this:

```hs
{-# LANGUAGE KindSignatures #-}
{-# LANGUAGE MultiParamTypeClasses #-}
{-# OPTIONS_GHC -ddump-splices #-}
module Bug where

import Data.Singletons.TH

$(singletons [d|
  infixl 5 `C`
  class C a b
  |])
```

You'll get the following code:

```
Bug.hs:(9,3)-(12,4): Splicing declarations
    singletons
      [d| infixl 5 `C_aejo`

          class C_aejo a_aejp b_aejq |]
  ======>
    infixl 5 `C_aejS`
    class C_aejS a_aejT b_aejU
    class PC (a_aejT :: GHC.Types.Type) (b_aejU :: GHC.Types.Type)
    infixl 5 `SC`
    class SC a_aejT b_aejU
```

Notice that there's an `infixl` declaration for `SC`, but not `PC`.

RyanGlScott:
Issue #412 revealed so many bugs in how `singletons` handles fixity declarations that it drove me to write `Note [singletons and fixity declarations]` to document that various wisdom one needs to handle them correctly:

* `singInfixDecl` now returns a `Maybe`, since the only type constructors that need to be singled are classes. `promoteInfixDecl` now also makes a similar special case when dealing with type constructors.
* The `promoteReifiedInfixDecls`/`singReifiedInfixDecls` functions are now used in the code that promotes/singles classes to ensure that their fixity declarations will be handled even if `genSingletons` is used instead of the `singletons` functions.

Fixes #412.

### DIFF
--- a/src/Data/Singletons/Single/Fixity.hs
+++ b/src/Data/Singletons/Single/Fixity.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE ScopedTypeVariables #-}
 module Data.Singletons.Single.Fixity where
 
 import Prelude hiding ( exp )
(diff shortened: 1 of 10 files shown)


src/Data/Singletons/Single/Fixity.hs
```diff
@@ -7,27 +8,128 @@ import Data.Singletons.Util
...
+
+  infixl 5 `FooSym0`
+  infixl 5 `FooSym1`
+  ...
+
+-----
+-- Wrinkle 1: When not to promote/single fixity declarations
+-----
+
+Rules are meant to be broken, and the general rule above is no exception. There
```
goldfirere: At least you haven't broken the rule that rules are meant to be broken.

PR #427, opened 2019-12-21
Introduce Options and OptionsMonad

### LINKED ISSUE #204: Allow custom naming hooks for autogenerated code
I ran into yet another naming conflict when using `singletons` recently:

```haskell
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE TypeInType #-}
module Bug where

import Data.Singletons.TH

$(singletons [d|
      data Ratio1 a = a :%  a
      data Ratio2 a = a :%% a
    |])
```

Compiling this with `singletons-2.3` yields the following (abridged) `-ddump-splices` output and error message:

```
GHCi, version 8.2.0.20170523: http://www.haskell.org/ghc/  :? for help
Bug.hs:(10,3)-(13,6): Splicing declarations
    singletons
      [d| data Ratio1_a3DY a_a3E1 = a_a3E1 :%_a3DZ a_a3E1
          data Ratio2_a3DW a_a3E0 = a_a3E0 :%%_a3DX a_a3E0 |]
  ======>
    data Ratio1_a6oe a_a6oi = a_a6oi :%_a6of a_a6oi
    data Ratio2_a6og a_a6oj = a_a6oj :%%_a6oh a_a6oj
    data instance Sing (z_a6oA :: Ratio1_a6oe a_a6oi)
      = forall (n_a6oB :: a_a6oi)
               (n_a6oC :: a_a6oi). z_a6oA ~ (:%_a6of) n_a6oB n_a6oC =>
        (:%%) (Sing (n_a6oB :: a_a6oi)) (Sing (n_a6oC :: a_a6oi))
    data instance Sing (z_a6oJ :: Ratio2_a6og a_a6oj)
      = forall (n_a6oK :: a_a6oj)
               (n_a6oL :: a_a6oj). z_a6oJ ~ (:%%_a6oh) n_a6oK n_a6oL =>
        (:%%%) (Sing (n_a6oK :: a_a6oj)) (Sing (n_a6oL :: a_a6oj))

Bug.hs:10:3: error:
    Multiple declarations of ‘:%%’
    Declared at: Bug.hs:10:3
                 Bug.hs:10:3
   |
10 | $(singletons [d|
   |   ^^^^^^^^^^^^^^...
```

Urk. We have a conflict between the `Ratio2` constructor and the singleton constructor for `Ratio1`. We could conceivably change the rules for generating singleton constructor names so as to avoid this problem, but I struggle to think of something that wouldn't create the potential for ambiguity somewhere else.

Instead of finagling the name generation rules, a better solution would be to allow users finer control over what gets generated. In the [spirit of `Data.Aeson.TH`](http://hackage.haskell.org/package/aeson-1.2.1.0/docs/Data-Aeson-TH.html#t:Options), we could offer an `Options` datatype (name subject to bikeshedding) that bundles hooks for generating `Name`s. Something like:

```haskell
data Options = Options
  { typeOptions :: TypeOptions
  , functionOptions :: FunctionOptions
  }

data DatatypeOrClass = IsDatatype | IsClass

data TypeOptions = TypeOptions
  { promotedKind :: DatatypeOrClass -> Name -> Name
  , singletonType :: DatatypeOrClass -> Name -> Name
  }

data ConstructorOrValue = IsConstructor | IsValue

data FunctionOptions = FunctionOptions
  { promotedType :: ConstructorOrValue -> Name -> Name
  , singletonFunction :: ConstructorOrValue -> Name -> Name
  , defunctionalizationSymbol :: ConstructorOrValue -> Int -> Name -> Name
  }
```

And then we could provide variants of `singletons`, `promote`, etc. that take `Options` as arguments.

RyanGlScott:
This patch introduces an `Options` data type and an `mtl`-like `OptionsMonad` class for monads that carry `Options`. At present, the only things one can do with `Options` are:

* Toggle the generation of `SingKind` instances. Suppressing `SingKind` instances provides an effective workaround for #150.
* Hook into the TH machinery's naming conventions for promoted and singled names. This fixes #204.

The vast majority of this patch simply adds plumbing by using `OptionsMonad` in places that need it. See the `D.S.TH.Options` module for where most of the new code is housed, as well as the `T150` and `T204` test cases for examples of how to use it.

### DIFF
--- a/src/Data/Singletons/TH/Options.hs
+++ b/src/Data/Singletons/TH/Options.hs
@@ -1,0 +1,259 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE TemplateHaskell #-}
+
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Data.Singletons.TH.Options
+-- Copyright   :  (C) 2019 Ryan Scott
+-- License     :  BSD-style (see LICENSE)
+-- Maintainer  :  Ryan Scott
+-- Stability   :  experimental
+-- Portability :  non-portable
+--
+-- This module defines 'Options' that control finer details of how the Template
+-- Haskell machinery works, as well as an @mtl@-like 'OptionsMonad' class
+-- and an 'OptionsM' monad transformer.
+--
+----------------------------------------------------------------------------
+
+module Data.Singletons.TH.Options
+  ( -- * Options
+    Options, defaultOptions
+    -- ** Options record selectors
+  , genSingKindInsts
+  , promotedClassName
+  , promotedValueName
+  , singledDataTypeName
+  , singledClassName
+  , singledDataConName
+  , singledValueName
+  , defunctionalizedName
+    -- ** Derived functions over Options
+  , promotedTopLevelValueName
+  , promotedLetBoundValueName
+  , defunctionalizedName0
+
+    -- * OptionsMonad
+  , OptionsMonad(..), OptionsM, withOptions
+  ) where
+
+import Control.Applicative
+import Control.Monad.IO.Class (MonadIO)
+import Control.Monad.Reader (ReaderT(..), ask)
+import Control.Monad.RWS (RWST)
+import Control.Monad.State (StateT)
+import Control.Monad.Trans.Class (MonadTrans(..))
+import Control.Monad.Writer (WriterT)
+import Data.Singletons.Names ( consName, listName, nilName
+                             , mk_name_tc, mkTupleDataName, mkTupleTypeName
+                             , sconsName, sListName, snilName
+                             , splitUnderscores
+                             )
+import Data.Singletons.Util
+import Language.Haskell.TH.Desugar
+import Language.Haskell.TH.Syntax hiding (Lift(..))
+
+-- | Options that control the finer details of how @singletons@' Template
+-- Haskell machinery works.
+data Options = Options
+  { genSingKindInsts        :: Bool
+    -- ^ If 'True', then 'SingKind' instances will be generated. If 'False',
+    --   they will be omitted entirely. This can be useful in scenarios where
+    --   TH-generated 'SingKind' instances do not typecheck (for instance,
+    --   when generating singletons for GADTs).
+  , promotedClassName    :: Name -> Name
+    -- ^ Given the name of the original, unrefined class, produces the name of
+    --   the promoted equivalent of the class.
+  , promotedValueName    :: Name -> Maybe Uniq -> Name
+    -- ^ Given the name of the original, unrefined value, produces the name of
+    --   the promoted equivalent of the value. This is used for both top-level
+    --   and @let@-bound names, and the difference is encoded in the
+    --   @'Maybe' 'Uniq'@ argument. If promoting a top-level name, the argument
+    --   is 'Nothing'. If promoting a @let@-bound name, the argument is
+    --   @Just uniq@, where @uniq@ is a globally unique number that can be used
+    --   to distinguish the name from other local definitions of the same name
+    --   (e.g., if two functions both use @let x = ... in x@).
+  , singledDataTypeName  :: Name -> Name
+    -- ^ Given the name of the original, unrefined data type, produces the name
+    --   of the corresponding singleton type.
+  , singledClassName     :: Name -> Name
+    -- ^ Given the name of the original, unrefined class, produces the name of
+    --   the singled equivalent of the class.
+  , singledDataConName   :: Name -> Name
+    -- ^ Given the name of the original, unrefined data constructor, produces
+    --   the name of the corresponding singleton data constructor.
+  , singledValueName     :: Name -> Name
+    -- ^ Given the name of the original, unrefined value, produces the name of
+    --   the singled equivalent of the value.
+  , defunctionalizedName :: Name -> Int -> Name
+    -- ^ Given the original name and the number of parameters it is applied to
+    --   (the 'Int' argument), produces a type-level function name that can be
+    --   partially applied when given the same number of parameters.
+    --
+    --   Note that defunctionalization works over both term-level names
+    --   (producing symbols for the promoted name) and type-level names
+    --   (producing symbols directly for the name itself). As a result, this
+    --   callback is used for names in both the term and type namespaces.
+  }
+
+-- | Sensible default 'Options'.
+--
+-- 'genSingKindInsts' defaults to 'True'.
+-- That is, 'SingKind' instances are generated.
+--
+-- The default behaviors for 'promotedClassName', 'promotedValueNamePrefix',
+-- 'singledDataTypeName', 'singledClassName', 'singledDataConName',
+-- 'singledValueName', and 'defunctionalizedName' are described in the
+-- \"On names\" section of the @singletons@ @README@.
+defaultOptions :: Options
+defaultOptions = Options
+  { genSingKindInsts     = True
+  , promotedClassName    = promoteClassName
+  , promotedValueName    = promoteValNameLhs
+  , singledDataTypeName  = singTyConName
+  , singledClassName     = singClassName
+  , singledDataConName   = singDataConName
+  , singledValueName     = singValName
+  , defunctionalizedName = promoteTySym
+  }
+
+-- | Given the name of the original, unrefined, top-level value, produces the
+-- name of the promoted equivalent of the value.
+promotedTopLevelValueName :: Options -> Name -> Name
+promotedTopLevelValueName opts name = promotedValueName opts name Nothing
+
+-- | Given the name of the original, unrefined, @let@-bound value and its
+-- globally unique number, produces the name of the promoted equivalent of the
+-- value.
+promotedLetBoundValueName :: Options -> Name -> Uniq -> Name
+promotedLetBoundValueName opts name = promotedValueName opts name . Just
+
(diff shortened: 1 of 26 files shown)


src/Data/Singletons/TH/Options.hs
```diff
@@ -0,0 +1,223 @@
...
+-- Haskell machinery works.
+data Options = Options
+  { genSingKindInsts        :: Bool
+    -- ^ If 'True', then 'SingKind' instances will be generated. If 'False',
+    --   they will be omitted entirely. This can be useful in scenarios where
+    --   TH-generated 'SingKind' instances do not typecheck (for instance,
+    --   when generating singletons for GADTs).
+  , promotedClassName       :: Name -> Name
+    -- ^ Controls how promoted type classes are named.
+  , promotedValueNamePrefix :: (String, String) -> Name -> Name
```
goldfirere: This is a bit strange, in that the arguments tell you what prefixes you want. Could some specific `Options` ignore them? Looking at use cases (quickly), it looks like maybe instead you want a `Bool` that says whether the name will be `let`-bound or `lambda`-bound.

src/Data/Singletons/TH/Options.hs
```diff
@@ -0,0 +1,223 @@
...
+                             , sconsName, sListName, snilName
+                             , splitUnderscores
+                             )
+import Data.Singletons.Util
+import Language.Haskell.TH.Desugar
+import Language.Haskell.TH.Syntax hiding (Lift(..))
+
+-- | Options that control the finer details of how @singletons@' Template
+-- Haskell machinery works.
+data Options = Options
```
goldfirere: Either here or with `defaultOptions`, the commentary should explain how these functions work by giving the canonical example.

src/Data/Singletons/TH/Options.hs
```diff
@@ -0,0 +1,223 @@
...
+-- | Declare the 'Options' that a TH computation should use.
+withOptions :: Options -> OptionsM m a -> m a
+withOptions opts (OptionsM x) = runReaderT x opts
+
+-- used when a value name appears in a pattern context
+-- works only for proper variables (lower-case names)
+-- the (String, String) is a pair of prefixes to prepend, where the first
+-- element is used for alphanumeric names and the second element is used for
+-- symbolic names.
+promoteValNameLhsPrefix :: (String, String) -> Name -> Name
```
goldfirere: This, too, is a bit awkward with the prefixes.

src/Data/Singletons/TH/Options.hs
RyanGlScott: Indeed, I think the only prefix that ever gets added is `"LetXYZ"`, where `XYZ` is some unique number generated by `qNewUnique`. We could conceivably just hard-wire that prefix into the default implementation of `promotedValueNamePrefix`, but that would require making it monadic (due the need to use `qNewUnique`), unlike every other naming hook. Is this a sign that we should make every other naming hook monadic for consistency?

src/Data/Singletons/TH/Options.hs
RyanGlScott: Done. To avoid repeating information, I included a pointer to the "On names" section of the `singletons` `README` to describe the default behavior of `promotedClassName` and friends.

src/Data/Singletons/TH/Options.hs
RyanGlScott: I've just pushed a change which tries out this design:

```hs
  , promotedValueName    :: Name -> Maybe Uniq -> Name
    -- ^ Given the name of the original, unrefined value, produces the name of
    --   the promoted equivalent of the value. This is used for both top-level
    --   and @let@-bound names, and the difference is encoded in the
    --   @'Maybe' 'Uniq'@ argument. If promoting a top-level name, the argument
    --   is 'Nothing'. If promoting a @let@-bound name, the argument is
    --   @Just uniq@, where @uniq@ is a globally unique number that can be used
    --   to distinguish the name from other local definitions of the same name
    --   (e.g., if two functions both use @let x = ... in x@).
```

As well as the following derived functions:

```hs
promotedTopLevelValueName :: Options -> Name -> Name
promotedLetBoundValueName :: Options -> Name -> Uniq -> Name
```

How does this look?

src/Data/Singletons/TH/Options.hs
```diff
@@ -0,0 +1,255 @@
...
+    -- ^ Given the name of the original, unrefined class, produces the name of
+    --   the singled equivalent of the class.
+  , singledDataConName   :: Name -> Name
+    -- ^ Given the name of the original, unrefined data constructor, produces
+    --   the name of the corresponding singleton data constructor.
+  , singledValueName     :: Name -> Name
+    -- ^ Given the name of the original, unrefined value, produces the name of
+    --   the singled equivalent of the value.
+  , defunctionalizedName :: Name -> Int -> Name
+    -- ^ Given the original name of a function (term- or type-level) and the
```
goldfirere: Do we really need this to work on both promoted and unpromoted names? That seems gratuitously unfriendly. Actually, maybe I just don't understand this documentation. It says both "original" and "type-level". How do these relate? (This is honest confusion, not just prodding.) Maybe this is because `singletons` allows the defunctionalization of user-written type families? What about type and data constructors? This documentation suggests that it works only over functions.

src/Data/Singletons/TH/Options.hs
```diff
@@ -0,0 +1,255 @@
...
+data Options = Options
+  { genSingKindInsts        :: Bool
+    -- ^ If 'True', then 'SingKind' instances will be generated. If 'False',
+    --   they will be omitted entirely. This can be useful in scenarios where
+    --   TH-generated 'SingKind' instances do not typecheck (for instance,
+    --   when generating singletons for GADTs).
+  , promotedClassName    :: Name -> Name
+    -- ^ Given the name of the original, unrefined class, produces the name of
+    --   the promoted equivalent of the class.
+  , promotedValueName    :: Name -> Maybe Uniq -> Name
```
goldfirere: I like this design *much* more than the previous one. I haven't looked for knock-on effects, but if they are reasonable, then I think this is a nice way to do what you want.

src/Data/Singletons/TH/Options.hs
goldfirere: General theme is good -- I have a specific query elsewhere.

src/Data/Singletons/TH/Options.hs
RyanGlScott: > Maybe this is because singletons allows the defunctionalization of user-written type families?

Indeed, this is precisely the reason. Type synonyms can be defunctionalized too. (In fact, so can type classes, oddly enough. I've opened a separate issue about this at #429.)

> What about type and data constructors? This documentation suggests that it works only over functions.

Sort of: see #429 for the confusing state of affairs on what can be defunctionalized and what can't.

In any case, I don't think we need a precise answer to #429 yet in order to document `defunctionalizedName` better. I've revised the Haddocks for `defunctionalizedName` to remove the emphasis on functions and to explicitly mention that defunctionalization works on both term- and type-level names:

```hs
  , defunctionalizedName :: Name -> Int -> Name
    -- ^ Given the original name and the number of parameters it is applied to
    --   (the 'Int' argument), produces a type-level function name that can be
    --   partially applied when given the same number of parameters.
    --
    --   Note that defunctionalization works over both term-level names
    --   (producing symbols for the promoted name) and type-level names
    --   (producing symbols directly for the name itself). As a result, this
    --   callback is used for names in both the term and type namespaces.
```

Does this look better to you?

src/Data/Singletons/TH/Options.hs
RyanGlScott: > I haven't looked for knock-on effects, but if they are reasonable, then I think this is a nice way to do what you want.

There are no knock-on _downsides_ to the best of my knowledge. There may be a knock-on _benefit_ in that this patch fixes a subtle bug related to how local fixity declarations are promoted. The bugfix is contained in this part of the diff:

```diff
@@ -615,20 +629,19 @@ promoteInfixDecl name fixity = do
     -- Certain function names do not change when promoted (e.g., infix names),
     -- so don't bother with them.
     promote_infix_val :: q (Maybe DDec)
-    promote_infix_val
-      | nameBase name == nameBase promoted_name
-      = never_mind
-      | otherwise
-      = finish promoted_name
-
-    promoted_name :: Name
-    promoted_name = promoteValNameLhs name
+    promote_infix_val = do
+      opts <- getOptions
+      let promoted_name :: Name
+          promoted_name = promotedValueName opts name mb_let_uniq
+      if nameBase name == nameBase promoted_name
+         then never_mind
+         else finish promoted_name
```

Notice that the old code used `promoteValNameLhs` (the previous name for `promotedTopLevelName`), but this is just plain wrong in the presence of local fixity declarations like this one:

```hs
f :: Bool
f = g () () where
  infixl 4 `g`
  g :: () -> () -> Bool
  g _ _ = True
```

Interestingly, this bug can't be observed in today's `singletons` due to [this GHC bug](https://gitlab.haskell.org/ghc/ghc/issues/17608) that causes all TH-quoted local fixity declarations to be dropped. Once that bug is fixed (which [should be soon](https://gitlab.haskell.org/ghc/ghc/merge_requests/2372)), then the code above would error if singled. However, this patch solves the problem by taking the `let`-bound unique (`mb_let_uniq`) into account when promoting the names in a fixity declaration.

src/Data/Singletons/TH/Options.hs
goldfirere: Yes -- great.

PR #432, opened 2020-01-26
The great StandaloneKindSignatures patch of 2020

RyanGlScott:
This is a mega-patch that overhauls `singletons` to use `StandaloneKindSignatures` as a means to:

1. Order kind variables more precisely (as a step towards #378).
2. Greatly simplify the algorithm used for defunctionalization.

This is separated into two separate commits. Here is the first commit message:

    Make TH generate standalone kind signatures
    
    This patch addresses chunk (2) in
    https://github.com/goldfirere/singletons/issues/378#issuecomment-528519679
    by making the TH machinery generate standalone kind signatures
    wherever possible, inching closer to the goal of preserving the order
    of `forall`'s type variables in all promoted and singled code. I say
    "inching" since this post does not quite get there (see the "Caveats"
    section at the bottom of this commit message), but it comes very
    close.
    
    In brief, the TH machinery now generates SAKS for the following:
    
    * Promoted top-level functions
    
      See `D.S.Promote.promoteLetDecRHS` for how this is implemented.
    
    * Defunctionalization symbols
    
      SAKS provide a wonderful opportunity to clean up the implementation
      in `D.S.Promote.Defun`, which is currently somewhat crufty. The
      code used to jump through hoops in order to give
      defunctionalization symbols CUSKs, but now that we live in a
      post-CUSKs world, we can get rid of all the ad hoc-ness in
      `Note [Defunctionalization and dependent quantification]`.
      Instead, we now use a much simpler algorithm that only depends
      on whether the thing being defunctionalized has a SAK or not.
      See `Note [Defunctionalization game plan]` in `D.S.Promote.Defun`.
    
    * Singled data type declarations
    
      If a data type's order of kind variables is known (e.g., if it was
      declared with a SAK), then the singled version of the data type
      will use a SAK to preserve the same kind variable order. See the
      code in `D.S.Single.Data.singDataD`.
    
    * Promoted and singled class declarations
    
      Similarly, if the order of kind variables is known for a class `C`,
      then the TH machinery will generate SAKS for `PC` and `SC` to
      preserve the same kind variable order. See the code in
      `D.S.Promote.promoteClassDec` and `D.S.Single.singClassD`.
    
    Naturally, lots of test cases' expected outputs change as a result of
    this patch. See `Singletons.T378{a,b}` for some new (or altered)
    cases that test new code paths.
    
    -------------
    -- Caveats --
    -------------
    
    Alas, not everything preserves the order of type variables after this
    patch. Some notable exceptions are:
    
    * `let`- or `where`-bound functions
    
      When promoting this function:
    
      ```hs
      f :: Bool
      f = let x = True
              g :: () -> Bool
              g _ = x
          in g ()
      ```
    
      The TH machinery will give the promoted `F` type family a SAK, but
      not the `G` type family. See
      `Note [No SAKs for let-decs with local variables]` in `D.S.Promote`
      for an explanation.
    
    * Promoted class methods
    
      This class:
    
      ```hs
      class C (b :: Type) where
        m :: forall a. a -> b -> a
      ```
    
      Is promoted like so:
    
      ```hs
      class PC (b :: Type) where
        type M (x :: a) (y :: b) :: a
      ```
    
      The order of the variables `a` and `b` are different between the
      type of `m` and the kind of `M`. See
      `Note [Promoted class methods and kind variable ordering]` in
      `D.S.Promote`.
    
    * Fully saturated defunctionalization symbols
    
      If you defunctionalize the `Id` type family, you'll generate two
      defunctionalization symbols:
    
      ```hs
      type IdSym0 :: forall a. a ~> a
      data IdSym0 f where ...
    
      type IdSym1 (x :: a) = Id x :: a
      ```
    
      Notice that unlike `IdSym0`, `IdSym1` (the "fully saturated"
      defunctionalization symbol) does _not_ have a SAK. This is because
      in general, giving fully saturated defunctionalization symbols SAKS
      can lead to kind errors. See
      `Note [No SAKs for fully saturated defunctionalization symbols]` in
      `D.S.Promote.Defun` for the sordid story.

And here is the second commit message:

    Make TH generate standalone kind signatures
    
    This patch addresses chunk (2) in
    https://github.com/goldfirere/singletons/issues/378#issuecomment-528519679
    by making the TH machinery generate standalone kind signatures
    wherever possible, inching closer to the goal of preserving the order
    of `forall`'s type variables in all promoted and singled code. I say
    "inching" since this post does not quite get there (see the "Caveats"
    section at the bottom of this commit message), but it comes very
    close.
    
    In brief, the TH machinery now generates SAKS for the following:
    
    * Promoted top-level functions
    
      See `D.S.Promote.promoteLetDecRHS` for how this is implemented.
    
    * Defunctionalization symbols
    
      SAKS provide a wonderful opportunity to clean up the implementation
      in `D.S.Promote.Defun`, which is currently somewhat crufty. The
      code used to jump through hoops in order to give
      defunctionalization symbols CUSKs, but now that we live in a
      post-CUSKs world, we can get rid of all the ad hoc-ness in
      `Note [Defunctionalization and dependent quantification]`.
      Instead, we now use a much simpler algorithm that only depends
      on whether the thing being defunctionalized has a SAK or not.
      See `Note [Defunctionalization game plan]` in `D.S.Promote.Defun`.


src/Data/Singletons/Promote.hs
```diff
@@ -816,6 +924,69 @@ promoteLetDecRHS rhs_sort type_env fix_env mb_let_uniq name let_dec_rhs = do
...
+
+  Alas, GHC currently does not allow wildcards in SAKs. See GHC#17432.
+
+* Use visible dependent quantification to avoid having to say what the kind
+  of `x` is:
+
+    type G :: forall x -> () -> Bool
+
+  A clever trick to be sure, but it doesn't quite do what we want, since
+  GHC will generalize that kind to become `forall (x :: k) -> () -> Bool`,
```
int-index: It's worse than that, you'd also run into #17072

RyanGlScott: That's true, although GHC#17072 is unlikely to be an issue in this particular circumstance. The promoted versions of local functions are, by virtue of being local, an internal implementation detail, so we are free to choose whatever kinds we want for them. There may be other problems with using SAKS here besides excessive kind generalization that I haven't thought of, but I would be surprised if GHC#17072 were one of them.

src/Data/Singletons/Partition.hs
```diff
@@ -49,17 +49,23 @@ data PartitionedDecs =
         , pd_ty_syn_decs :: [TySynDecl]
         , pd_open_type_family_decs :: [OpenTypeFamilyDecl]
         , pd_closed_type_family_decs :: [ClosedTypeFamilyDecl]
+        , pd_standalone_kind_sigs :: Map Name DKind
```
RyanGlScott: I'm split on this part of the patch. I added this in an effort to make reifying locally declared SAKS slightly more efficient. I'm not sure that the amount of code that we have to add to support this (see all the call sites of `dsReifyTypeWith`) is worth it. Perhaps effort would be better spent making local reification more efficient on `th-desugar`'s end? After all, locally reifying something in `th-desugar` basically amounts to scanning through a giant `[Dec]`.

RyanGlScott: I've filed an issue about the time complexity of local reification in goldfirere/th-desugar#130. In the meantime, I'll do a quick microbenchmark and see how long it takes to compile `singletons` with and without the `pd_standalone_kind_sigs` hack. If the difference is negligible, I think I'll just rip out the hack.

src/Data/Singletons/Promote.hs
```diff
@@ -750,53 +815,96 @@ promoteLetDecRHS rhs_sort type_env fix_env mb_let_uniq name let_dec_rhs = do
...
+    promote_let_dec_rhs all_locals m_ldrki ty_num_args
                         promote_thing mk_prom_eqns mk_alet_dec_rhs = do
       opts <- getOptions
-      tyvarNames <- mapM (const $ qNewName "a") m_argKs
+      tyvarNames <- replicateM ty_num_args (qNewName "a")
       let proName         = promotedValueName opts name mb_let_uniq
           local_tvbs      = map DPlainTV all_locals
-          m_fixity        = OMap.lookup name fix_env
-          args            = zipWith inferMaybeKindTV tyvarNames m_argKs
+          args            = case m_ldrki of
```
goldfirere: Too many matches on `Maybe`s here. Could this just be done once, with `(a, b, c, d) | Just (LDRKI ...) <-... = ... | otherwise = ...`?

src/Data/Singletons/Promote/Defun.hs
```diff
@@ -340,279 +386,496 @@ defunctionalize name m_fixity m_arg_tvbs' m_res_kind' = do
...
+The catch is that the kind of QuuxSym0, `forall k. Type ~> k ~> Type`, is
+slightly more general than it ought to be. In practice, however, this is
+unlikely to be a problem as long as you apply QuuxSym0 to arguments of the
+right kinds.
 
-       To continue the Symmetry example:
+Note [No SAKs for fully saturated defunctionalization symbols]
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+When generating defunctionalization symbols, most of the symbols are data
+types. The last one, however, is a type family. For example, this code:
```
goldfirere: Do you mean that the last one is a type *synonym*?

src/Data/Singletons/Promote/Defun.hs
```diff
@@ -340,279 +386,496 @@ defunctionalize name m_fixity m_arg_tvbs' m_res_kind' = do
...
-       (b) We flatten this into a list of well scoped type variables:
+  type ConstSym2 (x :: a) (y :: b) = Const x y :: a
 
-             [ k
-             , (t :: k)
-             , (a :: Proxy t)
-             , (y :: Proxy t)
-             , (e :: (a :: Proxy (t :: k)) :~: (y :~: Proxy (t :: k)))
-             ]
+ConstSym2, the sole type family of the bunch, is what is referred to as a
```
goldfirere: s/family/synonym/

src/Data/Singletons/Promote/Defun.hs
```diff
@@ -340,279 +386,496 @@ defunctionalize name m_fixity m_arg_tvbs' m_res_kind' = do
...
-dropTvbKind (DKindedTV n _)   = DPlainTV n
+However, the kind of (*>) is
+`forall (f :: Type -> Type) (a :: Type) (b :: Type). f a -> f b -> f b`.
+This is not general enough for (*>@#@$$$), which expects kind-polymorphic `f`,
+`a`, and `b`, leading to a kind error. You might think that we could somehow
+infer this information, but note the quoted definition of Applicative (and
+PApplicative, as a consequence) omits the kinds of `f`, `a`, and `b` entirely.
+Unless we were to implement full-blown kind inference inside of Template
+Haskell (which is a tall order), the kind `f a -> f b -> f b` is about as good
+as we can get.
```
goldfirere: Why doesn't this problem happen for unsaturated symbols? I don't see how saturation causes a difference here. Or maybe it does happen for unsaturated symbols, but no one notices...

src/Data/Singletons/Promote/Defun.hs
```diff
@@ -340,279 +386,496 @@ defunctionalize name m_fixity m_arg_tvbs' m_res_kind' = do
...
+
+    natMinus :: Nat -> Nat -> Nat
+    natMinus Z       _     = Z
+    natMinus (S a)   (S b) = natMinus a b
+    natMinus a@(S _) Z     = a
+    |])
+
+But I would generally prefer to avoid having the user add extraneous pattern
+matches when possible. Given the choice between expressiveness and SAKs, I give
+the edge to expressiveness.
```
goldfirere: But we explicitly don't support overlapping equations, and the killer example uses those. I don't necessarily think it's worth fixing this now, but if the lack of a SAK here causes problems, I think this is viable.

CHANGES.md
```diff
@@ -24,9 +24,14 @@ Changelog for singletons project
...
-  constructors. See the `Support for TypeApplications` section of the `README`
-  for more details.
+  variables in type signatures during promotion and singling. See the
+  `Support for TypeApplications` section of the `README` for more details.
+
+  When generating type-level declarations in particular (e.g., promoted type
+  families or defunctionalization symbols), `singletons` will likely also
+  generate standalone kind signatures to preserve type variable order. As a
+  result, most `singletons` code that uses Template Haskell will require the
+  use of the `StandaloneKindSignatures` extension to work.
```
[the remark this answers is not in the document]
RyanGlScott: Yes, there are various parts of `singletons` that now assume the behavior of `NoCUSKs` (e.g., the new defunctionalization algorithm no longer factors in CUSKs). I have not tested what would happen if you try combining `StandaloneKindSignatures` and `CUSKs` with `singletons`-generated code, but I have a funny feeling that it wouldn't end well. In light of this, we should just explicitly state that `singletons` requires `NoCUSKs`. I've updated the changelog and `README` to do just that.

src/Data/Singletons/Partition.hs
RyanGlScott: The results of my microbenchmark confirm my suspicions: the `pd_standalone_kind_sigs` hack isn't pulling its weight. Here is the time to compile `singletons` with the hack:

```
real    3m30.704s
user    3m23.989s
sys     0m2.987s
```

And without the hack:

```
real    3m27.500s
user    3m24.402s
sys     0m2.839s
```

I've ripped out the hack in my latest update. Let's pursue optimizing the speed of `dsReifyType` in goldfirere/th-desugar#130 instead.

src/Data/Singletons/Prelude/Monad/Internal.hs
```diff
@@ -44,16 +45,6 @@ import Data.Singletons.Prelude.Base
 import Data.Singletons.Prelude.Instances
 import Data.Singletons.Single
 
-{-
```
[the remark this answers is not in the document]
RyanGlScott: Actually, this Note has been obsolete ever since #416 (which did indeed delete various kind annotations), but I somehow skipped over this Note at the time. I decided to posthumously remove it in an act of "while-I'm-in-town" cleanup.

src/Data/Singletons/Promote.hs
```diff
@@ -387,10 +444,11 @@ promoteInstanceDec orig_meth_sigs cls_tvbs_map
                                   , id_sigs     = inst_sigs
                                   , id_meths    = meths }) = do
   opts <- getOptions
-  cls_tvb_names <- lookup_cls_tvb_names
+  cls_tvbs <- lookup_cls_tvbs
```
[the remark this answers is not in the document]
RyanGlScott: I originally thought that I would need to use the class-bound `TyVarBndr`s in the SAKS for promoted instance methods, so these changes were made in pursuit of that goal. However, I later realized that the SAKS for promoted instance methods don't need to explicitly bind kind variables at all—see my comment starting with "`For class method RHS helper functions, don't bother quantifying`..." in `promote_let_dec_ty`. As a result, this `TyVarBndr` refactoring went unused.

I still kind of like the new code over the old code, since we now only invoke `map extractTvbName` once instead of twice. But if you'd prefer I can revert this change.

src/Data/Singletons/Promote.hs
```diff
@@ -532,28 +590,29 @@ promoteMethod meth_sort orig_sigs_map (meth_name, meth_rhs) = do
...
               -- We don't have an InstanceSig, so we must compute the kind to use
               -- ourselves.
-              (arg_kis, res_ki) <- lookup_meth_ty
+              (_, arg_kis, res_ki) <- lookup_meth_ty
               -- Substitute for the class variables in the method's type.
               -- See Note [Promoted class method kinds]
               let arg_kis' = map (substKind cls_subst) arg_kis
                   res_ki'  = substKind cls_subst res_ki
-              pure (arg_kis', res_ki')
+                  tvbs'    = toposortTyVarsOf (arg_kis' ++ [res_ki'])
```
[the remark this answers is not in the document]
RyanGlScott: Dadgummit, I could have swore that I cross-referenced `Note [Promoted class methods and kind variable ordering]` here. I made sure to do so in my latest update.

src/Data/Singletons/Promote.hs
```diff
@@ -563,9 +622,10 @@ promoteMethod meth_sort orig_sigs_map (meth_name, meth_rhs) = do
                      -- See Note [Using dsReifyTypeNameInfo when promoting instances]
           case mb_info of
             Just (DTyConI (DOpenTypeFamilyD (DTypeFamilyHead _ tvbs mb_res_ki _)) _)
-              -> let arg_kis = map (defaultToTypeKind . extractTvbKind) tvbs
-                     res_ki  = defaultToTypeKind (resultSigToMaybeKind mb_res_ki)
-                  in pure (arg_kis, res_ki)
+              -> let arg_kis = map (defaultMaybeToTypeKind . extractTvbKind) tvbs
+                     res_ki  = defaultMaybeToTypeKind (resultSigToMaybeKind mb_res_ki)
+                     tvbs'   = toposortTyVarsOf (arg_kis ++ [res_ki])
```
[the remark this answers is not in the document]
RyanGlScott: Dadgummit again. Fixed.

src/Data/Singletons/Promote.hs
```diff
@@ -705,20 +770,20 @@ promoteLetDecRHS :: LetDecRHSSort
                  -> Name                 -- name of the thing being promoted
                  -> ULetDecRHS           -- body of the thing
                  -> PrM ( DDec          -- "type family"
+                        , Maybe DDec    -- standalone kind signature for type family
```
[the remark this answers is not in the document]
RyanGlScott: Good idea. Done.

src/Data/Singletons/Promote.hs
RyanGlScott: Funny enough, I chose this design because I thought `(a, b, c, d) | Just (LDRKI ...) <-... = ... | otherwise = ...` looked uglier than the current design (which, admittedly, is not much prettier in my eyes). I've changed the code back to your suggested design—is this to your liking?

src/Data/Singletons/Promote/Defun.hs
RyanGlScott: Eep. Yes, I did. Fixed.

src/Data/Singletons/Promote/Defun.hs
RyanGlScott: Eep again. Fixed.

src/Data/Singletons/Promote/Defun.hs
RyanGlScott: The difference is that unsaturated defunctionalization symbols are implemented as GADTs, and GADTs let you match on kind arguments. For instance, in the definition of `(*>@#@$)`:

```hs
type (*>@#@$) :: f a ~> f b ~> f b
data (*>@#@$) f where
  (*>@#@$###) :: forall f z. SameKind (Apply (*>@#@$) z) ((*>@#@$$) z) =>
                        (*>@#@$) f
```

It's subtle, but the use of `SameKind` in `(*>@#@$###)` causes `k` to become equal to `Type` (where `k` is the kind of `a` and `b`). This is similar to the reason why defining unsaturated defunctionalization symbols as type families makes this work out (`natMinus` issues notwithstanding). I made a mention of that later in the Note, but it's worth pointing out here as well. I've updated the prose to do just that.

src/Data/Singletons/Promote/Defun.hs
RyanGlScott: The lack of a SAK (hey, that rhymes!) certainly doesn't cause problems now, but it does create a major hurdle if we ever want to fully support promoting visible type applications. This is because `f` is always promoted to `FSym0`, even if it isn't applied to any arguments. In order to promote `f @x` to `FSym0 @x`, it is imperative that the order of kind variables in `FSym0` is the same as in `f`.

Moreover, this isn't even the only hurdle we'd have to clear in order to promote visible type application. See also `Note [Promoted class methods and kind variable ordering]` and `Note [No SAKs for let-decs with local variables]` in this patch. These problems discussed in those Notes don't have a clear solution, which leads me to believe that fixing chunk (3) in https://github.com/goldfirere/singletons/issues/378#issuecomment-528519679 is a pipe dream (at least for now).

In light of all this, it seems silly to change saturated defunctionalization symbols to type families (which induces a breaking change) when it won't even get us that much closer to the stated goal of promoting visible type applications. Let me know if you feel strongly otherwise, but I'm disinclined to rock the boat without first charting a clear path to shore.

src/Data/Singletons/Promote.hs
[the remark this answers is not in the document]
goldfirere: You can leave it changed. It was more that I was worried I was missing something critical.

src/Data/Singletons/Promote.hs
goldfirere: Maybe I'm wrong, but I think we can do even better. There are several other definitions that immediately scrutinize `Maybe`s. Like `m_sak_dec`, `defun_ki`, and `tf_head`.

src/Data/Singletons/Promote/Defun.hs
goldfirere: What's worse than a mixed metaphor? A pseudo-mixed one, which you successfully pulled off in your last sentence. I'll have to remember that one.

We're in agreement here. My comment was merely to say that we should not unduly prioritize accepting this example.

src/Data/Singletons/Promote.hs
RyanGlScott: I've factored out the definitions of `mk_sak_dec`, `defun_ki`, and `tf_head` to use a single `case`.

goldfirere: What we really have here is three cases:
1. We have no information.
2. We have a `LetDecRHSKindInfo`, but no SAK.
3. We have a `LetDecRHSKindInfo` with a SAK.

The three cases lead to slightly different algorithms. But this fact is not currently all that apparent from the code. My hope is that there would be one `case`, with three branches, that distinguishes the three cases. That said, perhaps that would cause other trouble in the code. I don't want this to hold up the patch -- just trying to make the code paths more apparent.

RyanGlScott: Ah, now I see what you're getting at. Indeed, that is a nice way to structure this code, and thinking about it that way made me realize that some of the things returned from the first tuple-returning `case` only exist to be fed into the second tuple-returning `case`. Combining both these `case`s into one allowed me to reduce the number of tupled return values from seven to four, which is much nicer.

I took some liberties with your suggestion and numbered the cases as 1/2(a)/2(b) instead of 1/2/3, since the last two cases share a few commonalities (e.g., they compute the exact same `lde_kvs_to_bind`). How does this look now?

RyanGlScott: I haven't head any objections by now, so I'll interpret that to mean that there are no further suggestions to this code. Let's land this.

goldfirere: I hadn't had time to look over the past few days, but yes I'm quite happy with this now. Thanks!

Issue #433, opened 2020-02-05
Promoting expression signatures can fail on GHC 8.10+


RyanGlScott:
`singletons` used to have no trouble promoting this code:

```hs
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE GADTs #-}
{-# LANGUAGE PolyKinds #-}
{-# LANGUAGE ScopedTypeVariables #-}
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE TypeApplications #-}
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE UndecidableInstances #-}
module Bug where

import Data.Singletons.Prelude
import Data.Singletons.TH

$(singletons [d|
  f :: forall a. a -> a
  f x = id (x :: a)
  |])
```

However, if you try compiling this with GHC 8.10, you'll get the following error:

```
[1 of 1] Compiling Bug              ( Bug.hs, interpreted )

Bug.hs:14:3: error:
    The exact Name ‘a_a83H’ is not in scope
      Probable cause: you used a unique Template Haskell name (NameU), 
      perhaps via newName, but did not bind it
      If that's it, then -ddump-splices might be useful
   |
14 | $(singletons [d|
   |   ^^^^^^^^^^^^^^...
```

Big yikes. What changed? Inspecting the `-ddump-simpl` reveals that this is how `f` is promoted:

```hs
    type family F (a_a83K :: a_a83H) :: a_a83H where
      F x_a83M = Apply IdSym0 (x_a83M :: a_a83H)
```

Note that on GHC 8.10 or later, the `a_a83H` in `Apply IdSym0 (x_a83M :: a_a83H)` is out of scope due to [GHC Proposal 103](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0103-no-kind-vars.rst) having been implemented. Unfortunately, this presents another incongruity between terms and types that `singletons` must bridge, since not even giving `F` a standalone kind signature would bring `a_a83H` into scope.

One possibile way forward here is to use visible kind applications to bind kind variables. That is, promote `F` like this instead:

```hs
type family F (x :: a) :: a where
  F @a x = Apply IdSym0 (x :: a)
```

This works in that particular example, but there is danger on the horizon. What about this code?

```hs
$(singletons [d|
  g :: forall a. a -> a
  g (x :: b) = id (x :: b)
  |])
```

[...]

PR #436, opened 2020-02-10
Single record selectors as top-level functions

### LINKED ISSUE #364: Singled records can behave counterintuitively
For the longest time, something about the way record selectors are singled has felt slightly off to me. I think the best way to illustrate the unease I feel is to point to this example:

```haskell
newtype Unit = MkUnit { runUnit :: () }
-- runUnit :: Unit -> ()

f :: Unit -> ()
f = runUnit
```

Surprisingly, this code successfully promotes, but doesn't single: `sF` will fail to typecheck. What is going on here? The heart of the matter is that this is the definition of `sRunUnit`:

```haskell
data instance Sing :: Unit -> Type where
  SMkUnit :: { sRunUnit :: Sing u } -> Sing (MkUnit u)
```

In other words,

```haskell
sRunUnit :: Sing (MkUnit u) -> Sing u
```

For the rest of this issue, I will refer to this style of singling record selectors as the "traditional" approach. This approach is actually quite distinct from how most definitions are singled. For instance, compare `sRunUnit` to an alternative way to define a "record selector" for `MkUnit` (I put "record selector" in quotes because it's not really one, but it's as close as I can get without using record syntax):

```haskell
runUnit2 :: Unit -> ()
runUnit2 (MkUnit unit) = unit
```

When singled, this becomes:

```haskell
sRunUnit2 :: Sing (unit :: Unit) -> Sing (RunUnit2 unit)
```

For the rest of this issue, I will refer to this as the "manual" approach to singling record selectors. Unlike the traditional approach, the manual approach does not constrain the type of its argument (beyond saying that it should be of kind `Unit`), and it incorporates a type family into its return type.

What's even more interesting is that using the traditional `runUnit` causes `f = runUnit` to fail to single, but swapping `runUnit` out the its manual equivalent, `runUnit2`, makes `f` single successfully. That's because in the traditional definition of `f`:

```haskell
sF :: Sing (unit :: Unit) -> Sing (F unit)
sF = sRunUnit
```

There is no constraint in scope that `unit ~ MkUnit u` (for some `u`), so `sRunUnit` fails to typecheck. On the other hand, `sRunUnit2` requires no such constraint, so it happily typechecks.

To make things even stranger, the traditional approach to singling records is somewhat at odds with the way records are promoted. Here is the promoted version of `runUnit`:

```haskell
-- type RunUnit :: Unit -> ()
type family RunUnit (unit :: Unit) :: () where
  RunUnit (MkUnit unit) = unit
```

In spirit, this is actually closer to the manual approach than the traditional one!

That's not to say that the traditional approach is totally useless. While I've demonstrated some scenarios where the manual approach wins out, there are other situations where the traditional approach is better:

1. Only the traditional definition can be used with record syntax (record construction, record updates, pattern matching, etc.).
2. In some scenarios, traditional singled records can avoid partiality. If you were to single the following data type:

   ```haskell
   data Vec :: Type -> Nat -> Type where
     VNil :: Vec a Z
     (:#) :: { vhead :: a, vtail :: Vec a n } -> Vec a (S n)
   ```

   Then if `sVhead` and `sVtail` were defined traditionally, they would be total functions, as their type signatures would require that the first argument have type `Sing (x :# xs)` (i.e., passing `SVNil` would be a type error). If they were defined manually, however, they would be partial.

This seems to indicate to me that perhaps `singletons` should be offering _both_ forms of records. If nothing else, `singletons`' own code would benefit from having the manual versions available, as there are many times where we have to jump through hoops to single code which uses newtype record selectors. If we did this, though, we'd have to answer the following questions:

1. What should the naming conventions for traditional and manual singled record selectors be?
2. If the naming convention for manual records ends up being different from what `singletons` currently offers, should we change the naming conventions for promoted records to match that of manual singled records? (If we don't, we'd have a discrepancy between the two, despite the fact that manual singled records would use these promoted records in their return types.)

Another option is to just observe this weirdness in the `README`, and require that anyone who wants to write functions like `f` above will have to define manual record selectors, well, manually. Currently, the `README` is rather silent about this point—the only oddity about records that it makes note of concerns record updates, which is a rather different problem.

RyanGlScott:
`singletons` has traditionally singled record selectors "in-place".
For example, `data T = MkT { unT :: Bool }` would be singled as
`data ST :: T -> Type where SMkT :: { sUnT :: Sing b } -> ST (MkT b)`.
This may seem like a sensible choice, but it has some unfortunate
consequences:

* This function will not typecheck when singled:

  ```hs
  f :: T -> Bool
  f = unT
  ```

  This is because the type of `sUnT` is `Sing (MkT b) -> b`, which
  is not general enough for the type of `sF`, which is
  `Sing (t :: T) -> Sing (F t :: Bool)`.
* It is impossible to single a data type with multiple constructors
  that share a record name, since each occurrence of a record
  selector in a data type is required to have the same type.

For these reasons and more discussed in
`Note [singletons and record selectors]` in `D.S.Single.Data`, we
have decided in #364 to single record selectors as simple top-level
functions. That is, we would generate the following for `sUnT`:

```hs
data ST :: T -> Type where
  SMkT :: Sing b -> ST (MkT b)

sUnT :: Sing (t :: T) -> Sing (UnT t :: Bool)
sUnT (MkT x) = x
```

This brings the treatment of singled record selectors in line with
promoted record selectors (note that `UnT` is also a top-level type
family) and avoids the drawbacks mentioned above. The drawback is
that it is no longer possible to use record syntax in combination
with `SMkT`, although record selectors for singleton data constructors
are already quite buggy (see
https://github.com/goldfirere/singletons/issues/364#issuecomment-582476156),
so this is arguably not that huge of a loss.

This change allows `D.S.Prelude.*` to single code that is much closer
to the original code found in `base`. As one example, the changes in
`D.S.Prelude.Foldable` should give a pretty good idea of the kind of
code that can now be singled.

Fixes #364.

### DIFF
--- a/src/Data/Singletons/Prelude/Foldable.hs
+++ b/src/Data/Singletons/Prelude/Foldable.hs
@@ -102,7 +102,7 @@ import Control.Monad
 import Data.Kind
 import Data.List.NonEmpty (NonEmpty(..))
 import Data.Monoid hiding (All(..), Any(..), Endo(..), Product(..), Sum(..))
-import qualified Data.Monoid as Monoid (All(..), Any(..), Product(..), Sum(..))
+import qualified Data.Monoid as Monoid (Product(..), Sum(..))
 import Data.Singletons.Internal
 import Data.Singletons.Prelude.Base
   hiding (Foldr, FoldrSym0, FoldrSym1, FoldrSym2, FoldrSym3, sFoldr)
@@ -127,11 +127,11 @@ import Data.Singletons.Prelude.Semigroup.Internal
   hiding ( AllSym0(..),     AllSym1,     SAll
          , AnySym0(..),     AnySym1,     SAny
          , FirstSym0,       FirstSym1,   SFirst
+         , GetFirstSym0,    sGetFirst
          , LastSym0,        LastSym1,    SLast
          , ProductSym0(..), ProductSym1, SProduct
          , SumSym0(..),     SumSym1,     SSum )
 import Data.Singletons.Prelude.Semigroup.Internal.Disambiguation
-import Data.Singletons.Promote
 import Data.Singletons.Single
 import Data.Singletons.TypeLits.Internal
 
@@ -146,6 +146,9 @@ data EndoSym0 tf
 type instance Apply EndoSym0 x = 'Endo x
 
 $(singletonsOnly [d|
+  appEndo :: Endo a -> (a -> a)
+  appEndo (Endo x) = x
+
   instance Semigroup (Endo a) where
           Endo x <> Endo y = Endo (x . y)
 
@@ -153,21 +156,10 @@ $(singletonsOnly [d|
           mempty = Endo id
   |])
 
-type MaxInternal :: Type -> Type
-newtype MaxInternal a = MaxInternal (Maybe a)
-type SMaxInternal :: MaxInternal a -> Type
-data SMaxInternal mi where
-  SMaxInternal :: Sing x -> SMaxInternal ('MaxInternal x)
-type instance Sing = SMaxInternal
-$(genDefunSymbols [''MaxInternal])
-
-type MinInternal :: Type -> Type
-newtype MinInternal a = MinInternal (Maybe a)
-type SMinInternal :: MinInternal a -> Type
-data SMinInternal mi where
-  SMinInternal :: Sing x -> SMinInternal ('MinInternal x)
-type instance Sing = SMinInternal
-$(genDefunSymbols [''MinInternal])
+$(singletons [d|
+  newtype MaxInternal a = MaxInternal { getMaxInternal :: Maybe a }
+  newtype MinInternal a = MinInternal { getMinInternal :: Maybe a }
+  |])
 
 $(singletonsOnly [d|
   instance Ord a => Semigroup (MaxInternal a) where
@@ -266,8 +258,7 @@ $(singletonsOnly [d|
       -- @foldr f z = 'List.foldr' f z . 'toList'@
       --
       foldr :: (a -> b -> b) -> b -> t a -> b
-      foldr f z t = case foldMap (Endo . f) t of
-                      Endo g -> g z
+      foldr f z t = appEndo (foldMap (Endo . f) t) z
 
       -- -| Right-associative fold of a structure, but with strict application of
       -- the operator.
@@ -302,8 +293,7 @@ $(singletonsOnly [d|
       -- @foldl f z = 'List.foldl' f z . 'toList'@
       --
       foldl :: (b -> a -> b) -> b -> t a -> b
-      foldl f z t = case foldMap (Dual . Endo . flip f) t of
-                      Dual (Endo g) -> g z
+      foldl f z t = appEndo (getDual (foldMap (Dual . Endo . flip f) t)) z
       -- There's no point mucking around with coercions here,
       -- because flip forces us to build a new function anyway.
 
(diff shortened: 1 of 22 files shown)


src/Data/Singletons/Prelude/Foldable.hs
```diff
@@ -370,26 +360,28 @@ $(singletonsOnly [d|
 
       -- -| The largest element of a non-empty structure.
       maximum :: forall a . Ord a => t a -> a
-      maximum x =
-        case foldMap (MaxInternal . Just) x of
-          MaxInternal y -> fromMaybe (errorWithoutStackTrace "maximum: empty structure") y
+      maximum = fromMaybe (errorWithoutStackTrace "maximum: empty structure") .
```
RyanGlScott: Ideally, I would have written this:

```hs
      maximum = fromMaybe (errorWithoutStackTrace "maximum: empty structure") .
       getMaxInternal . foldMap (MaxInternal . (Just :: a -> Maybe))
```

As that is closer to how `maximum` is actually defined in `base`. Sadly, this won't work in `singletons` due to #433.

src/Data/Singletons/Promote.hs
```diff
@@ -313,10 +305,21 @@ promoteLetDecs mb_let_uniq decls = do
...
 --
 --  * for each nullary data constructor we generate a type synonym
-promoteDataDec :: DataDecl -> PrM ()
-promoteDataDec (DataDecl _name _tvbs ctors) = do
-  ctorSyms <- buildDefunSymsDataD ctors
-  emitDecs ctorSyms
+--
+--  * we create a "manual" record selector function and return its DLetDecs so
+--    that they can be promoted later.
+--    See Note [singletons and record selectors] in D.S.Single.Data.
```
goldfirere: So this emits declarations for the promoted fixity declarations for record selectors, but just returns the record selector simulated functions unpromoted? If that simplifies the plumbing, then sure, but it should be carefully commented.

RyanGlScott: You raise a good point: the name "`promoteDataDec`" doesn't actually promote a data declaration (not even its record selectors), which makes its name rather misleading. We should name it something else.

Moreover, I'm not sure that `promoteDataDec` is the best place to put that call to `promoteReifiedInfixDecls`, since it would make more sense to put it in `promoteLetDecRHS` (which already does 90% of record selector promotion). Perhaps we should do the remaining 10% there as well.

I'll experiment with these ideas and push a new design when it's ready.

RyanGlScott: Hm, the name "`promoteDataDec`" isn't completely misleading after all, since it also defunctionalizes the data constructors.

Moreover, my plan of moving the invocation of `promoteReifiedInfixDecls` into `promoteLetDecRHS` won't work out cleanly, since `promoteLetDecRHS` is also on each class method whenever an instance is promoted. When promoting an instance method, the `Name` passed to `promoteLetDecRHS` happens to be the name of the method itself. If that name happens to have a fixity declaration in the module that defined it, then `promoteReifiedInfixDecls` will generate an _additional_ promoted fixity declaration every single time that method is defined in an instance! Ugh. We could imagine adding a special case in `promoteLetDecRHS` whenever we are dealing with an instance, but that solution doesn't seem much better than just keeping `promoteReifiedInfixDecls` in `promoteDataDec`.

Bottom line: I have decided to keep the name "`promoteDataDec`" and keep its implementation just the way it is. I have rewritten the documentation for `promoteDataDec`, both to clarify why it returns an unpromoted `DLetDec` and because some of its documentation was irrelevant (e.g., it mentions deriving `Eq` instances, even though `promoteDataDec` has nothing to do with that).

PR #437, opened 2020-02-10
Add Data.Singletons.Prelude.Proxy

### LINKED ISSUE #435: Constrained singleton generation
Assuming I have the following class with associated type family and singleton
member function, is it possible to generate a type family and singleton which
depends on these? For example:

```haskell
class Barable (a :: Type) where
  type Bar a :: Nat -> Type
  sBar :: Proxy a -> Sing (n :: Nat) -> Sing (Bar a)

$(singletonsOnly [d|
  foo :: Type -> Nat -> [Type]
  foo a n = if n == 0
             then []
             else bar a n : foo a (n - 1)
  |])
```

Given the above, or something similar, I'd like to have generated the equivalent
of the following:

```haskell
-- A type family for `foo`
type family Foo (a :: Type) (n :: Nat) :: [Type] where
  Foo a 0 = '[]
  Foo a n = Bar a n ': Foo a (n - 1)

-- A singleton for `foo`. Importantly not taking a `Sing a` which may not
-- exist. Either of the following function or equivalent would do. The first of
-- which has an explicit `Barable a` constraint paired with `Proxy a`, the
-- second uses a GADT `SomeBar a` to transport the `Barable a` witness.

sFoo :: Barable a => Proxy a -> Sing (n :: Nat) -> Sing [Type]
sFoo a n = case n %== (sing @0) of
  STrue  -> SNil
  SFalse -> sBar a n `SCons` sFoo a (n %- sing @1)

sFoo :: forall a n. SomeBar a -> Sing (n :: Nat) -> Sing [Type]
sFoo a@SomeBar n = case n %== (sing @0) of
  STrue  -> SNil
  SFalse -> sBar (Proxy @a) n `SCons` sFoo a (n %- sing @1)
```

I'd really like to write `foo`/`Foo`/`sFoo` once and have `singletons` generate
the type level and term versions. As far as I can tell though, `singletons`
doesn't support such a thing, the particular stumbling blocks with the approach
above are:

- Expressing the `Barable a` constraint on the term `a` in `foo a n`
- Passing `a` not as `Sing a` in `sFoo`

I played around a bit with wrapping `a` in `foo` with some newtype with a
`Sing` instance which contained a `Barable a` witness without success.

RyanGlScott:
Even in a library as deeply entrenched in type-level trickery as
`singletons` is, there are still some situations that call for
proxies. See, for example, the use case in #435, which uses not only
the classic `Proxy` data type, but also the /singled/ version of
`Proxy`! The author of this use case requested that the singled
version of `Proxy` be made available in `singletons`, so this patch
accomplishes that.

Fixes #435.

### DIFF
--- a/src/Data/Singletons/Prelude/Traversable.hs
+++ b/src/Data/Singletons/Prelude/Traversable.hs
@@ -62,6 +62,7 @@ import Data.Singletons.Prelude.Identity
 import Data.Singletons.Prelude.Instances
 import Data.Singletons.Prelude.Monad.Internal
 import Data.Singletons.Prelude.Monoid
+import Data.Singletons.Prelude.Proxy
 import Data.Singletons.Single
 
 type StateL :: Type -> Type -> Type
@@ -204,7 +205,13 @@ $(singletonsOnly [d|
   deriving instance Traversable NonEmpty
   deriving instance Traversable (Either a)
   deriving instance Traversable ((,) a)
-  deriving instance Traversable (Const m)
+
+  instance Traversable Proxy where
+      traverse _ _ = pure Proxy
+      sequenceA _ = pure Proxy
+      mapM _ _ = pure Proxy
+      sequence _ = pure Proxy
+
   deriving instance Traversable Dual
   deriving instance Traversable Sum
   deriving instance Traversable Product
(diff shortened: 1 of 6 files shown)


src/Data/Singletons/Prelude/Traversable.hs
```diff
@@ -204,7 +205,13 @@ $(singletonsOnly [d|
   deriving instance Traversable NonEmpty
   deriving instance Traversable (Either a)
   deriving instance Traversable ((,) a)
-  deriving instance Traversable (Const m)
```
expipiplus1: Should this have been deleted?

RyanGlScott: Sigh, that definitely wasn't intended. I've added this back in ad4a8e90f342183b685ce3bd935ea5d1bf6ad6a3.

Issue #429, opened 2019-12-26
Knowing what is (and isn't) defunctionalized can be confusing


RyanGlScott:
While pondering the questions in https://github.com/goldfirere/singletons/pull/427#discussion_r361532779, I realized that trying to document the existing status quo regarding what language constructs are OK to defunctionalize is much, much trickier than I thought. The main source of trickiness is that `singletons` has an inconsistent ruleset: whether something is defunctionalized or not depends on whether it is quoted (i.e., `$(singletons [d| ... |])`) or reified (i.e., `$(genDefunSymbols [...])`):

# Quoted

## Functions

Legal. Just define the function in quotes, e.g.,

```hs
$(singletons [d|
  foo :: Bool -> Bool
  foo = ...
  |])
```

And `singletons` will produce `FooSym0` and `FooSym1`.

## Classes

Legal, but only in the sense that quoted classes will produce defunctionalization symbols for the promoted class methods and associated type families, e.g.,

```
λ> ; $(singletons [d| class C a where type T a :: Bool; m :: a -> a |])
<interactive>:6:5-67: Splicing declarations
    singletons
      [d| class C_aejd a_aejg where
            type T_aeje a_aejg :: Bool
            m_aejf :: a_aejg -> a_aejg |]
  ======>
    class C_aejX a_aek0 where
      type T_aejY a_aek0 :: Bool
      m_aejZ :: a_aek0 -> a_aek0
    type TSym1 ...
    data TSym0 ...
    type MSym1 ...
    data MSym0 ...
```

Defunctionalization symbols for the class itself are not produced.

## Class methods

Legal. See the "Classes" section.

## Type families and synonyms

Legal. Quoted type families and synonyms are defunctionalized as you would expect. See also the "Classes" section for how associated type families are handled.

## Data types

Legal, but only in the sense that quoted data types will produce defunctionalization symbols for the promoted record selectors and data constructors, e.g.,

```
λ> ; $(singletons [d| data D a = MkD { unD :: a } |])
<interactive>:8:5-49: Splicing declarations
    singletons [d| data D_aelz a_aelC = MkD_aelA {unD :: a_aelC} |]
  ======>
    data D_aema a_aemd = MkD_aemb {unD_aemc :: a_aemd}
    type UnDSym1 ...
    data UnDSym0 ...
    type MkDSym1 ...
    data MkDSym0 ...
```

Defunctionalization symbols for the data type itself are not produced.

## Data constructors

Legal. See the "Data types" section.

[...]

PR #441, opened 2020-02-23
More thoroughly document what is (and isn't) allowed in the README

RyanGlScott:
This patch fleshes out some more details about what `singletons`
can and can't do in its `README`. The key changes are:

1. There is a new "Promotion and partial application" section that
   explains what defunctionalization is in some amount of detail.
   There is also a new subsection that explains the limitations of
   the `genDefunSymbols` function that were observed in #429.
2. The "Supported Haskell constructs" section has received some more
   love. Some Haskell features were inaccurately characterized
   (e.g., pattern signatures are really only partially supported),
   so I also reorganized some of the bullet points. I have also added
   a new bullet point for `ScopedTypeVariables` under the
   "Little to no support" section, as #433 reveals that promoting
   functions that rely on the behavior of `ScopedTypeVariables` is
   terribly fragile (and not easy to fix).
3. Lots of little formatting and grammar fixes to make the prose in
   the `README` flow better.

Note that this patch does _not_ fix either of #429 or #433—it just
documents the rather unsatisfying current state of affairs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,12 @@ singletons 2.7
 [![Hackage](https://img.shields.io/hackage/v/singletons.svg)](http://hackage.haskell.org/package/singletons)
 [![Build Status](https://travis-ci.org/goldfirere/singletons.svg?branch=master)](https://travis-ci.org/goldfirere/singletons)
 
-This is the README file for the singletons library. This file contains all the
+This is the README file for the `singletons` library. This file contains all the
 documentation for the definitions and functions in the library.
 
-The singletons library was written by Richard Eisenberg, <rae@cs.brynmawr.edu>, and
-with significant contributions by Jan Stolarek, <jan.stolarek@p.lodz.pl>.  There
+The `singletons` library was written by Richard Eisenberg (<rae@cs.brynmawr.edu>) and
+with significant contributions by Jan Stolarek (<jan.stolarek@p.lodz.pl>) and
+Ryan Scott (<ryan.gl.scott@gmail.com>). There
 are two papers that describe the library. Original one, _Dependently typed
 programming with singletons_, is available
 [here](https://cs.brynmawr.edu/~rae/papers/2012/singletons/paper.pdf) and will
@@ -18,18 +19,19 @@ paper, _Promoting Functions to Type Families in Haskell_, is available
 and will be referenced in this documentation as the
 "promotion paper".
 
-Ryan Scott, <ryan.gl.scott@gmail.com>, is an active maintainer.
+Ryan Scott (<ryan.gl.scott@gmail.com>) is the active maintainer.
 
-Purpose of the singletons library
----------------------------------
+Purpose of the `singletons` library
+-----------------------------------
 
 The library contains a definition of _singleton types_, which allow programmers
 to use dependently typed techniques to enforce rich constraints among the types
 in their programs. See the singletons paper for a more thorough introduction.
 
 The package also allows _promotion_ of term-level functions to type-level
-equivalents. Accordingly, it exports a Prelude of promoted and singletonized
-functions, mirroring functions and datatypes found in Prelude, `Data.Bool`,
+equivalents and _singling_ functions to dependently typed equivalents.
+Accordingly, it exports a Prelude of promoted and singled
+functions, mirroring functions and datatypes found in the `Prelude`, `Data.Bool`,
 `Data.Maybe`, `Data.Either`, `Data.Tuple` and `Data.List`. See the promotion
 paper for a more thorough introduction.
 
@@ -40,7 +42,7 @@ knowledge of dependent types.
 Compatibility
 -------------
 
-The singletons library requires GHC 8.10.1 or greater. Any code that uses the
+The `singletons` library requires GHC 8.10.1 or greater. Any code that uses the
 singleton generation primitives needs to enable a long list of GHC
 extensions. This list includes, but is not necessarily limited to, the
 following:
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -4,10 +4,10 @@ singletons 2.7
 [![Hackage](https://img.shields.io/hackage/v/singletons.svg)](http://hackage.haskell.org/package/singletons)
 [![Build Status](https://travis-ci.org/goldfirere/singletons.svg?branch=master)](https://travis-ci.org/goldfirere/singletons)
 
-This is the README file for the singletons library. This file contains all the
+This is the README file for the `singletons` library. This file contains all the
 documentation for the definitions and functions in the library.
 
-The singletons library was written by Richard Eisenberg, <rae@cs.brynmawr.edu>, and
+The `singletons` library was written by Richard Eisenberg, <rae@cs.brynmawr.edu>, and
 with significant contributions by Jan Stolarek, <jan.stolarek@p.lodz.pl>.  There
```
goldfirere: Ryan's name should be in this list, for sure.

README.md
```diff
@@ -20,16 +20,17 @@ and will be referenced in this documentation as the
 
 Ryan Scott, <ryan.gl.scott@gmail.com>, is an active maintainer.
```
goldfirere:
```suggestion
Ryan Scott, <ryan.gl.scott@gmail.com>, is the active maintainer.
```

README.md
```diff
@@ -214,13 +202,13 @@ Equality classes
 There are two different notions of equality applicable to singletons: Boolean
 equality and propositional equality.
 
-* Boolean equality is implemented in the type family `(:==)` (which is actually
-a synonym for the type family `(==)` from `Data.Type.Equality`) and the class
-`SEq`. See the `Data.Singletons.Prelude.Eq` module for more information.
+* Boolean equality is implemented in the type family `(==)` (in the `PEq`
+  class) and the `(%==`) function class (in the `SEq` class).
```
goldfirere:
```suggestion
  class) and the `(%==)` method (in the `SEq` class).
```

README.md
```diff
@@ -327,30 +315,178 @@ $(promote [d|
...
+defunctionalization symbols will be generated:
+
+```hs
+type IdSym0 :: a ~> a
+type IdSym0 x = x
+
+type IdSym1 (x :: a) = Id a
+```
+
+In general, a function that accepts _N_ arguments generates _N_ defunctionalization
```
goldfirere:
```suggestion
In general, a function that accepts _N_ arguments generates _N+1_ defunctionalization
```

README.md
```diff
@@ -327,30 +315,178 @@ $(promote [d|
...
+type Apply :: (a ~> b) -> a -> b
+type family Apply f x
+```
+
+Every defunctionalization symbol comes with a corresponding `Apply` instance
+(except for fully saturated defunctionalization symbols). For instance, here
+is the `Apply` instance for `IdSym0`:
+
+```hs
+type instance Apply IdSym0 x = x
```
goldfirere: Is that really true? I expected `IdSym1 x` on the RHS.

README.md
```diff
@@ -327,30 +315,178 @@ $(promote [d|
...
+```
+
+The `(~>)` kind is used when promoting higher-order functions so that partially
+applied arguments can be passed to them. For instance, here is a better attempt
```
goldfirere: "better": Is this correct? Or merely one step better and we'll get even more better later?

README.md
```diff
@@ -327,30 +315,178 @@ $(promote [d|
...
+at promoting `map`:
+
+```hs
+type Map :: (a ~> b) -> [a] -> [b]
+type family Map f xs where
+  Map _ '[]    = '[]
+  Map f (x:xs) = Apply f x : Map f xs
+```
+
+Now `map id` can be promoted to `Map IdSym0`, which typechecks without issue.
```
goldfirere: That's not quite right, because `Map IdSym0` is undersaturated.

README.md
```diff
@@ -327,30 +315,178 @@ $(promote [d|
...
+type MyTypeFamily :: Nat -> Bool
+type family MyTypeFamily n
+
+$(genDefunSymbols [''MyTypeFamily])
+```
+
+This can be especially useful if `MyTypeFamily` needs to be implemented by
+hand. Be aware of the following design limitations of `genDefunSymbols`:
+
+* `genDefunSymbols` only works for type-level declarations. Namely, it only
```
goldfirere: I thought we were going to have a `genAllDefunSymbols` or some such that went deeper...

README.md
```diff
@@ -327,30 +315,178 @@ $(promote [d|
...
+
+Note that `MyProxySym0` is a bit more general than it ought to be, since
+there is no dependency between the first kind (`Type`) and the second kind
+(`k`). But this would require the ability to write something like:
+
+```hs
+type MyProxySym0 :: forall (k :: Type) ~> k ~> Type
+```
+
+Which currently isn't possible. So for the time being, the kind of
```
goldfirere:
```suggestion
which currently isn't possible. So for the time being, the kind of
```

README.md
```diff
@@ -598,31 +734,54 @@ The following constructs are fully supported:
...
-explicitly indicate one's intentions by using the `DerivingStrategies` extension.
+## Partial support
+
+The following constructs are partially supported:
 
-`singletons` fully supports the `anyclass` strategy as well as the `stock` strategy
-(at least, for the classes listed above). `singletons` does not support the
-`newtype` strategy, as there is not an equivalent of `coerce` at the type level.
+* `deriving`
+* higher-kinded type variables
```
goldfirere: There is no follow-on section on this.

README.md
```diff
@@ -598,31 +734,54 @@ The following constructs are fully supported:
...
+* finite arithmetic sequences
```
goldfirere: There is no follow-on section to this.

README.md
```diff
@@ -598,31 +734,54 @@ The following constructs are fully supported:
...
+* records
+* signatures (e.g., `(x :: Maybe a)`) in patterns
+* functional dependencies
```
goldfirere: There is no follow-on section to this.

README.md
```diff
@@ -598,31 +734,54 @@ The following constructs are fully supported:
...
+* type families
```
goldfirere: There is no follow-on section to this.

README.md
```diff
@@ -633,67 +792,153 @@ since `singletons` desugars away most uses of record syntax. On the other hand,
...
+```haskell
+filter :: (a -> Bool) -> [a] -> [a]
+filter _pred []    = []
+filter pred (x:xs)
+  | pred x         = x : filter pred xs
+  | otherwise      = filter pred xs
+```
+
+Overlap is caused by `otherwise` catch-all guard, which is always true and thus
+erlaps with `pred x` guard.
```
goldfirere:
```suggestion
overlaps with `pred x` guard.
```

README.md
RyanGlScott: Similarly, I think this is a reference to the "Known bugs" section, which has a bullet point about type families. I have moved this bullet point to the "Partial support" section.

PR #444, opened 2020-04-09
Implement Refun for refunctionalising a defunctionalisation symbol

infinity0:
I wanted to, at the type-level, do `fmap (\v -> (x -> v)) vv` for some fixed but parameterised `x`.

After some poring through the docs I came up with `Fmap (FlipSym2 (~>@#@$) x) vv`, however this has the wrong kind:

```haskell
Prelude
Data.Kind
Data.Singletons
Data.Singletons.Prelude.Function
Data.Singletons.Prelude.Functor
> :kind! Fmap (FlipSym2 (~>@#@$) Int) '[Bool]
Fmap (FlipSym2 (~>@#@$) Int) '[Bool] :: [*]
= '[TyFun Bool Int -> *]
```

I tried searching through the existing docs and code for what I needed but it seems it's not implemented, so here is a PR for it. It works for my use-case:

```haskell
> :kind! Fmap (Refun0 .@#@$$$ (FlipSym2 (~>@#@$) Int)) '[Bool]
Fmap (Refun0 .@#@$$$ (FlipSym2 (~>@#@$) Int)) '[Bool] :: [*]
= '[Bool -> Int]
```

Apologies if it already exists. I thought it should already exist, but couldn't find it. I even ran `egrep -R '=\s*\w+\s*->'` but couldn't find anything obvious.

### DIFF
--- a/src/Data/Singletons.hs
+++ b/src/Data/Singletons.hs
@@ -60,6 +60,7 @@ module Data.Singletons (
 
   -- ** Defunctionalization
   TyFun, type (~>),
+  Refun, Refun0, Refun1,
   TyCon1, TyCon2, TyCon3, TyCon4, TyCon5, TyCon6, TyCon7, TyCon8,
   TyCon, Apply, type (@@), ApplyTyCon, ApplyTyConAux1, ApplyTyConAux2,
 
--- a/src/Data/Singletons/Internal.hs
+++ b/src/Data/Singletons/Internal.hs
@@ -288,6 +288,14 @@ type (@@) :: (k1 ~> k2) -> k1 -> k2
 type a @@ b = Apply a b
 infixl 9 @@
 
+-- | Refunctionalise a defunctionalised type function back to its type.
+type family Refun f where
+   Refun (a ~> b) = a -> b
+
+data Refun0 :: Type ~> Type
+type instance Apply Refun0 f = Refun1 f
+type Refun1 f = Refun f
+
 -- | Workhorse for the 'TyCon1', etc., types. This can be used directly
 -- in place of any of the @TyConN@ types, but it will work only with
 -- /monomorphic/ types. When GHC#14645 is fixed, this should fully supersede


RyanGlScott:
I'm a bit confused. You say:

> I wanted to, at the type-level, do `fmap (\v -> (x -> v)) vv` for some fixed but parameterised `x`.

But in your example:

> ```
> > :kind! Fmap (Refun0 .@#@$$$ (FlipSym2 (~>@#@$) Int)) '[Bool]
> Fmap (Refun0 .@#@$$$ (FlipSym2 (~>@#@$) Int)) '[Bool] :: [*]
> = '[Bool -> Int]
> ```

It looks like you're actually doing `fmap (\v -> (v -> x)) vv` for some fixed `x`.

In any case, I don't think you need anything like `Refun` to do what you seek. The existing [`TyCon` family of types](https://hackage.haskell.org/package/singletons-2.7/docs/Data-Singletons.html#t:TyCon1) should work just as well. For example, here is your example, but with `TyCon2`:

```
λ> :kind! Fmap (FlipSym2 (TyCon2 (->)) Int) '[Bool]
Fmap (FlipSym2 (TyCon2 (->)) Int) '[Bool] :: [*]
= '[Bool -> Int]
```

Or, if you wanted the flip the order of the argument and result types:

```
λ> :kind! Fmap (TyCon1 ((->) Int)) '[Bool]
Fmap (TyCon1 ((->) Int)) '[Bool] :: [*]
= '[Int -> Bool]
```

infinity0:
Sorry yes that was a typo, I did mean `fmap (\v -> (v -> x)) vv` with `x` as the result type.

Thanks for the hint, I guess this sort of knowledge comes with experience with the library. Perhaps it would be worthwhile mentioning `TyCon2 (->)` in the documentation for `(~>)`? Something like:

~~~~
-- Note: `(~>)` is relevant for singleton's defunctionalisation encoding of
-- type-level functions. To manipulate kinds involving actual `(->)` on the
-- type level, one should use `TyCon2 (->)` which is the defunctionalisation
-- symbol for `(->)`, rather than using the defunctionalisation symbol for this
-- type which would be `(~>@#@$)`.
~~~~

It's easy to overlook `TyCon2` since its current doc (well, that for `TyCon1`) talks about converting from `(->)` into `(~>)`, rather than preserving `(->)` when defining (or attempting to define) encoded functions on it.

RyanGlScott:
I agree that `(~>)` could use some more documentation. How about this?

```hs
-- | Something of kind @a '~>' b@ is a defunctionalized type function that is
-- not necessarily generative or injective. Defunctionalized type functions
-- (also called \"defunctionalization symbols\") can be partially applied, even
-- if the original type function cannot be. For more information on how this
-- works, see the "Promotion and partial application" section of the
-- @<https://github.com/goldfirere/singletons/blob/master/README.md README>@.
--
-- Normal type-level arrows @(->)@ can be converted into defunctionalization
-- arrows @('~>')@ by the use of the 'TyCon' family of types. (Refer to the
-- Haddocks for 'TyCon1' to see an example of this in practice.) For this
-- reason, we do not make an effort to define defunctionalize symbols for most
-- type constructors of kind @a -> b@, as they can be used in defunctionalized
-- settings by simply applying @TyCon{N}@ with an appropriate @N@.
--
-- This includes the @(->)@ type constructor itself, which is of kind
-- @'Type' -> 'Type' -> 'Type'@. One can turn it into something of kind
-- @'Type' '~>' 'Type' '~>' 'Type'@ by writing @'TyCon2' (->)@, or something of kind
-- @'Type' -> 'Type' '~>' 'Type'@ by writing @'TyCon1' ((->) t)@ (where @t :: 'Type').
```

infinity0:
That looks great to me, yes!

infinity0:
copyediting: in the middle paragraph, `defunctionalize symbols` -> `defunctionalization symbols`

RyanGlScott:
I've added these Haddocks in a0653082c8f9f3f600215f2a5f4a938e922acccb, which supersedes this patch. Many thanks for the documentation suggestions!

PR #458, opened 2020-04-28
Make {P,S}Eq less magical, don't make DefaultEq the default

### LINKED ISSUE #457: Reconsider the default implementation of PEq.==
There is a surprising amount of special-casing in `singletons` just for the sake of the `Eq` class. Ultimately, the root of all the special-casing is this default implementation for the associated `(==)` type:

https://github.com/goldfirere/singletons/blob/c6648cd1b71304a2f00ed25c2a844e5e289dd92d/src/Data/Singletons/Prelude/Eq.hs#L41-L48

Where:

https://github.com/goldfirere/singletons/blob/c6648cd1b71304a2f00ed25c2a844e5e289dd92d/src/Data/Singletons/Prelude/Eq.hs#L61-L64

The problem with this default is that for a significant portion of singleton types, `DefaultEq` is a rather roundabout way of defining `(==)`. In order to reduce the first equation to `True`, you have to know that they are the exact same type. For instance, if you tried implement a `PEq` instance for `Maybe` that uses `DefaultEq`, then you wouldn't be able to implement an implementation of `SEq` by simple pattern matching and recursion:

```hs
instance PEq (Maybe a)
instance SEq (Maybe a) where
  SNothing %== SNothing = STrue
  SNothing %== SJust{}  = SFalse
  SJust{}  %== SNothing = SFalse
  SJust sx %== SJust sy = sx %== sy
```
```
error:
    • Could not deduce: (n == n1)
                        ~ Data.Singletons.Prelude.Eq.DefaultEq ('Just n) ('Just n1)
      from the context: a1 ~ 'Just n
        bound by a pattern with constructor:
                   SJust :: forall a (n :: a). Sing n -> SMaybe ('Just n),
                 in an equation for ‘%==’
        at Bug.hs:22:3-10
      or from: b ~ 'Just n1
        bound by a pattern with constructor:
                   SJust :: forall a (n :: a). Sing n -> SMaybe ('Just n),
                 in an equation for ‘%==’
        at Bug.hs:22:16-23
      Expected type: Sing (a1 == b)
        Actual type: Sing (n == n1)
    • In the expression: sx %== sy
      In an equation for ‘%==’: SJust sx %== SJust sy = sx %== sy
      In the instance declaration for ‘SEq (Maybe a)’
    • Relevant bindings include
        sy :: Sing n1 (bound at Bug.hs:22:22)
        sx :: Sing n (bound at Bug.hs:22:9)
        (%==) :: Sing a1 -> Sing b -> Sing (a1 == b)
          (bound at Bug.hs:19:12)
   |
22 |   SJust sx %== SJust sy = sx %== sy
   |                           ^^^^^^^^^
```

In fact, I can only think of three examples where `DefaultEq` is actually used in practice: the `PEq` instances for `Nat`, `Symbol`, and `TYPE rep`. The `SEq` instances defined on top of them, however, are very atypical for the `singletons` library, as they rely on `Typeable` and `unsafeCoerce` tricks. Most other instances, however, do not (and cannot) make use of `DefaultEq`, which makes me believe that we should change it.

My proposal: make `(==)`'s default implementation be `\x y -> not (x == y)`, just like it is in `base`. Doing so would let us simply promote and single `Eq` using the existing TH machinery, rather than forcing us to laboriously define `PEq` and `SEq` by hand.

RyanGlScott:
Currently, the `PEq` and `SEq` classes are implemented by hand in
`D.S.Prelude.Eq` due to the default implementation, `DefaultEq`. But
as discussed in #457, this default isn't actually wise for most
singleton types. This patch removes `DefaultEq` as the default, which
allows `PEq` and `SEq` to be generated with Template Haskell, just
like other classes in the `D.S.Prelude.*` module namespace. As a
result, this fixes #457.

This refactoring led me to discover some other unusual treatment of
`PEq` and `SEq`. Namely, derived instances of `PEq` and `SEq` are
very magical in that they are handled entirely outside of
`D.S.Partition.partitionDecs`. But there isn't much of a good reason
to do so—we can just as well generate a normal `Eq` instance, then
promote and single it using the existing TH machinery. The only
magical thing about deriving `Eq` is that it also produces
`SDecide`/`TestEquality`/`TestCoercion` instances, but that can be
done independently of `PEq` and `SEq`.

This patch also refactors the treatment of `deriving Eq` to remove
the special casing for `PEq` and `SEq`:

* The `D.S.Promote.Eq` and `D.S.Single.Eq` modules are now gone,
  having been subsumed by the new `D.S.Deriving.Eq` module, which
  generates an instance just like `singletons` would for any other
  `stock`-derivable class. What remains of `D.S.Single.Eq` is now
  located in `D.S.Single.Decide`, which only exists to generate
  instances of `SDecide`, `TestEquality`, and `TestCoercion`.
* `D.S.TH` had strange functions `singEqInstanceOnly` and
  `singEqInstancesOnly` that generated instances of `SEq`, but not
  `PEq`. No other `stock`-derivable class had `-Only` variants like
  this, and given that `singEqInstance{s}` already exist, they were
  practically useless. I decided to take the opportunity to remove
  these warts.


src/Data/Singletons/Deriving/Eq.hs
```diff
@@ -0,0 +1,62 @@
...
+        [DConP lname lpats, DConP rname rpats]
+        (allExp (zipWith (\l r -> foldExp (DVarE equalsName) [l, r])
+                         lvars rvars))
+  | otherwise =
+      pure $ DClause
+        [DConP lname (replicate lNumArgs DWildP),
+         DConP rname (replicate rNumArgs DWildP)]
+        (DConE falseName)
+  where
+    allExp :: [DExp] -> DExp
```
goldfirere, line 56: Can this be `andExp` please?

RyanGlScott: Done.

RyanGlScott:
> We're deriving very routing, non-singletons-specific instances -- and even with inferred constraints!

Almost, but there are some aspects to `D.S.Deriving.*` that are unique to `singletons`:

1. Derived `Eq` instances generate more cases in `singletons` than in GHC. In GHC, a derived `Eq` instance for `Bool` would look (roughly) like this:

   ```hs
   instance Eq Bool where
     True  == True  = True
     False == False = True
     _     == _     = False
   ```

   In `singletons`, however, a derived `Eq` instance would need to match on every possile pair of constructors:

   ```hs
   instance Eq Bool where
     True  == True  = True
     False == False = True
     True  == False = False
     False == True  = False
   ```

   This is ultimately due to the need to work around goldfirere/th-desugar#6, which is very much a `singletons`-specific issue.
2. While it's true that `D.S.Deriving.*` does infer constraints, the way it does so is rather ugly compared to GHC. `singletons` simply gathers up all of the field types, sticks the class name in front of each type, and makes that the context. This leads to gnarly looking instance contexts like this one:

   ```hs
   instance (Eq a, Eq [a]) => Eq [a] where ...
   ```

   This is acceptable for `singletons`' purposes, since users have long since accepted the fact that they'll need to enable `UndecidableInstances` in order to TH-generated code to typecheck. GHC's own `deriving`, on the other hand, makes an effort to avoid inferring exotic constraints like these.

FWIW, I do maintain a [`deriving-compat`](http://hackage.haskell.org/package/deriving-compat) library whose functionality largely overlaps with `D.S.Deriving.*`. I don't think that it would be a drop-in replacement _per se_, but it does allow others to hook into it at the very least.

PR #462, opened 2020-05-19
Add promotedDataTypeOrConName option for custom promotion

### LINKED ISSUE #450: Add promotedDataTypeName option to make working with arrows/literals easier
The current state of the art for promoting data types that contain arrow or literal types is rather shoddy. First, you have to declare a base type, like so:

```hs
newtype Age' n = MkAge n
```

Next, you have to explicitly create term- and type-level versions of this base type:

```hs
-- Term-level
type Age = Age' Numeric.Natural.Natural
-- Type-level
type PAge = Age' GHC.TypeNats.Nat

type SAge :: PAge -> Type
data SAge age where
  SMkAge :: Sing n -> SAge (MkAge n)
type instance Sing = SAge

instance SingKind PAge where
  type Demote PAge = Age
  ...
```

As shown in the `Sing` and `SingKind` instances above, if you want to actually _use_ these types, it is almost guaranteed that you won't be able to use `singletons`' TH machinery to do so. This is because `singletons` is not aware of the relationship between `Age` and `PAge`, so it cannot know to promote `Age` to `PAge`. As a result, you basically have no choice but to promote all of your `Age`-related code by hand. Bummer.

But things don't have to be this terrible. Now that `singletons` has the ability to [hook into how it generates names during promotion](https://github.com/goldfirere/singletons/pull/427), there is a viable way to teach `singletons` to promote `Age` to `PAge`. In particular, I propose doing the following:

1. Add a new field to `D.S.TH.Options.Options` called `promotedDataTypeName :: Name -> Name` (in the spirit of the existing `singledDataTypeName`). By default, `promotedDataTypeName = id` since `DataKinds` automatically promotes data type names. The interesting part comes next:
2. Declare `Age'`, `Age`, and `PAge` as above. However, we no longer need to write out the `Sing`, `SingKind`, _etc._ instances by hand. Instead, we can let the TH machinery do it by hooking into `promotedDataTypeName` like so:

```hs
$(let handleAge :: Name -> Name
      handleAge n
        | n == ''Age = ''PAge
        | otherwise  = n in
  
  withOptions defaultOptions{promotedDataTypeName = handleAge} $
  genSingletons [''Age])
```

Now, `singletons` will have the foresight to promote `Age` to `PAge` automatically. This should make using `Age`/`PAge` much less painful in practice.

RyanGlScott:
This allows teaching the Template Haskell machinery about data types that need to be promoted in special ways (e.g., promoting `Nat` to `Natural`). This makes working with such types much less painful than before, where one always had to resort to writing such code by hand.

Fixes #450.

### DIFF
--- a/src/Data/Singletons/TH/Options.hs
+++ b/src/Data/Singletons/TH/Options.hs
@@ -22,6 +22,7 @@ module Data.Singletons.TH.Options
     -- ** Options record selectors
   , genQuotedDecs
   , genSingKindInsts
+  , promotedDataTypeOrConName
   , promotedClassName
   , promotedValueName
   , singledDataTypeName
@@ -45,7 +46,7 @@ import Control.Monad.RWS (RWST)
 import Control.Monad.State (StateT)
 import Control.Monad.Trans.Class (MonadTrans(..))
 import Control.Monad.Writer (WriterT)
-import Data.Singletons.Names (consName, listName, nilName, splitUnderscores)
+import Data.Singletons.Names
 import Data.Singletons.Util
 import Language.Haskell.TH.Desugar
 import Language.Haskell.TH.Syntax hiding (Lift(..))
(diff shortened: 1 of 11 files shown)


src/Data/Singletons/TH/Options.hs
```diff
@@ -235,20 +258,36 @@ promoteTySym name sat
 promoteClassName :: Name -> Name
 promoteClassName = prefixName "P" "#"
 
+promoteDataTypeOrConName :: Name -> Name
+promoteDataTypeOrConName nm
+  | nm == typeRepName               = typeKindName
```
RyanGlScott, line 263: This special case (which promoted `TypeRep` to `Type`) was inherited from the old code in `D.S.Promote.Type.promoteType_NC`, which I copied over wholesale. But to be honest, I'm really not sure why this special case exists in the first place. It's not like we use `TypeRep` as the right-hand side of a `Demote` instance anywhere. And even if we did, I'm not sure that we'd want this to special case to always be in effect. (After all, we don't have special cases for `Natural`/`Text` always be in effect.)

I'm inclined to just remove this special case. What do you think?

RyanGlScott: `git blame` reveals that commit c1344d6 introduced this special case. Unfortunately, that commit doesn't really contain any insights as to _why_ it's special-cased.

goldfirere: Then remove it. I think this was a half-hearted attempt to make more types "just promote" -- but it clearly never worked.

PR #463, opened 2020-06-04
Split singletons into three packages

### LINKED ISSUE #420: Question: singletons-core?
Sorry if this is not the right place to ask!

Question:
Would it be desirable/possible to have a separate package with only the TH machinery for generating boilerplate? (+ general-purpose stuff like `Sing` `SingI` etc)

Motivation:
Currently it takes a lot of time (~10minutes) to compile the entire package, even if I'm not using any of the "builtin" singletonised stuff. My current solution is to just hand-write the little bits I actually need, but I can foresee it quickly becoming overwhelming.

Thanks!

RyanGlScott:
This splits `singletons` into three packages, as proposed in https://github.com/goldfirere/singletons/issues/420#issuecomment-627322533:

* The new `singletons` library is now a minimal library that only provides `Data.Singletons`, `Data.Singletons.Decide`, `Data.Singletons.Sigma`, and `Data.Singletons.ShowSing` (if compiled with GHC 8.6 or later). `singletons` now supports building GHCs back to GHC 8.0, as well as GHCJS.
* The `singletons-th` library defines Template Haskell functionality for promoting and singling term-level definitions, but but nothing else. This library continues to require the latest stable release of GHC.
* The `singletons-base` library defines promoted and singled versions of definitions from the `base` library, which includes everything under `Data.Singletons.Prelude.*`. This library continues to require the latest stable release of GHC.

Fixes #420.


phadej:
Thanks! I'll take a closer look soon.

I think that `singletons-base` could use `reexported-modules: ` for singetons and singletons-th modules. Am I right to say that their versions are still very tightly coupled, so it hard to anticipate a `singletons` release without `singletons-base` release. And that if there is a change in `Data.Singletons.TH` it won't be affecting `singletons-base` as well. (`reexported-modules` require you to depend on minor versions of the packages providing said modules, but I don't expect much churn to occur due that versus convenience for users who would just want `singletons-base`).

RyanGlScott:
> I think that `singletons-base` could use `reexported-modules: ` for singetons and singletons-th modules.

That is an interesting suggestion. Historically, I've been averse to using `reexported-modules` due to serious bugs in certain versions of `Cabal`. (See [this](https://gitlab.haskell.org/ghc/ghc/issues/12653) GHC bug report.) I _think_ those bugs are gone in modern `Cabal`s due to #3906, but I haven't checked in a long time. Do you know what the minimum `Cabal` version I would need to use in `singletons-th.cabal` and `singletons-base.cabal` would be if I did add `reexported-modules`?

> Am I right to say that their versions are still very tightly coupled

Yes, `singletons-th` is very tightly coupled with `singletons` because of re-exports and cascading effects from the latter's API affecting the former. (Similarly with `singletons-base`'s dependencies on `singletons` and `singletons-th`.)

> (`reexported-modules` require you to depend on minor versions of the packages providing said modules, but I don't expect much churn to occur due that versus convenience for users who would just want `singletons-base`).

True, although I think we'd want to depend on minor versions regardless. What is the best way to do so? `singletons == 3.0.*`?

phadej:
> `singletons == 3.0.*`

Yes.

> minimum `Cabal` version

`cabal-version: 1.24` should be good which coincidentally the one bundled with GHC-8.0.2. I haven't noticed any problems with `optics` which uses `reexported-modules`, and `cabal-version: 1.24`.

singletons-base/CHANGES.md
```diff
@@ -0,0 +1,41 @@
...
+    promoting and singling term-level definitions, but but nothing else. This
+    library continues to require the latest stable release of GHC.
+  * The `singletons-base` library defines promoted and singled versions of
+    definitions from the `base` library, which includes everything under
+    `Data.Singletons.Prelude.*`. This library continues to require the latest
+    stable release of GHC.
+
+  Consult the changelogs for `singletons` and `singletons-th` for changes
+  specific to those libraries. For more information on this split, see the
+  [relevant GitHub discussion]()https://github.com/goldfirere/singletons/issues/420.
```
googleson78, line 22: Link seems broken

singletons-th/CHANGES.md
```diff
@@ -0,0 +1,105 @@
...
+    promoting and singling term-level definitions, but but nothing else. This
+    library continues to require the latest stable release of GHC.
+  * The `singletons-base` library defines promoted and singled versions of
+    definitions from the `base` library, which includes everything under
+    `Data.Singletons.Prelude.*`. This library continues to require the latest
+    stable release of GHC.
+
+  Consult the changelogs for `singletons` and `singletons-base` for changes
+  specific to those libraries. For more information on this split, see the
+  [relevant GitHub discussion]()https://github.com/goldfirere/singletons/issues/420.
```
googleson78, line 22: Here also

singletons/CHANGES.md
```diff
@@ -1,92 +1,25 @@
...
+    promoting and singling term-level definitions, but but nothing else. This
+    library continues to require the latest stable release of GHC.
+  * The `singletons-base` library defines promoted and singled versions of
+    definitions from the `base` library, which includes everything under
+    `Data.Singletons.Prelude.*`. This library continues to require the latest
+    stable release of GHC.
+
+  Consult the changelogs for `singletons-th` and `singletons-base` for changes
+  specific to those libraries. For more information on this split, see the
+  [relevant GitHub discussion]()https://github.com/goldfirere/singletons/issues/420.
```
googleson78, line 22: Ditto

phadej:
```diff
-cabal-version:  >= 1.24
+cabal-version:  >= 1.24
```

`cabal check` is your friend.

phadej:
I'm quite happy

I didn't go all the way to define the defunctionalized symbols as in `singletons-base`, but I don't see why I couldn't. 

```diff
--- a/src/Data/Singletons/Bool.hs
+++ b/src/Data/Singletons/Bool.hs
@@ -5,6 +5,8 @@
 {-# LANGUAGE RankNTypes          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeOperators       #-}
+{-# LANGUAGE TypeFamilies        #-}
+{-# LANGUAGE FlexibleInstances   #-}
 #if __GLASGOW_HASKELL__ >= 800
 {-# OPTIONS_GHC -Wno-redundant-constraints #-}
 #endif
@@ -43,6 +45,8 @@ import           Unsafe.Coerce      (unsafeCoerce)
 
 import Data.Proxy (Proxy (..))
 
+import qualified Data.Singletons as S
+
 -- $setup
 -- >>> :set -XDataKinds -XTypeOperators
 -- >>> import Data.Type.Dec (decShow)
@@ -68,6 +72,18 @@ instance Eq (SBool b) where
 instance Ord (SBool b) where
     compare _ _ = EQ
 
+-------------------------------------------------------------------------------
+-- Singletons
+-------------------------------------------------------------------------------
+
+type instance S.Sing = SBool
+instance SBoolI b => S.SingI (b :: Bool) where
+    sing = sbool
+instance S.SingKind Bool where
+    type Demote Bool = Bool
+    fromSing   = fromSBool
+    toSing   b = withSomeSBool b S.SomeSing
+
 -------------------------------------------------------------------------------
 -- conversion to and from explicit SBool values
 -------------------------------------------------------------------------------
```

i.e. it would take maybe an hour to make `singleton-bool` a superset of  `Data.Singletons.Prelude.Bool`

RyanGlScott:
Thanks, @phadej. I wonder if you have any feedback on using `singletons-th` as well? That was the one part of the proposal that was the shakiest, so it would be interesting to see how people experience using it in practice.

k0001:
@RyanGlScott once the new minimal `singletons` is out, I will definitely use some TH on top of it to accomplish what [this function](https://hackage.haskell.org/package/singlethongs-0.1/docs/Singlethongs.html#v:singlethongs) does, but using `SDecide` rather than `TestEquality`, seeing as that will be available in the minimal `singletons`. I was thinking of simply changing `singlethongs` to embrace the minimal `singletons`, but if you guys would prefer to keep that code as part of this repo as `singletons-th-lite`, then that's even better. I'm happy to deprecate  `singlethongs`. My main concern is that the code continues to compile on GHCJS as time goes by. GHCJS is currently more or less compatible with GHC 8.6.x.

Ideally, what I would like to be able to support using the minimal `singletons` _plus_ `singlethons-th-lite`/`singlethongs`, is the entirety of [`exinst`](https://hackage.haskell.org/package/exinst-0.7/docs/Exinst.html). However, there are some things there, such as [`Generic` instances for dependent pairs](https://github.com/k0001/exinst/blob/556035943d2a76d3cf796672313a948a0fe13787/exinst/lib/Exinst/Base.hs) that make use of `PEnum`, `PBounded`, which don't seem to be present in this proposed minimal `singletons` By the way, in `exinst`, these dependent pairs have the names `Some1`, `Some2`, … This is from before `Sigma` was in `singletons`. Would it be possible to export `PEnum`, `PBounded` from the new minimal `singletons` or that eventual `singlethons-th-lite`?

RyanGlScott:
> My main concern is that the code continues to compile on GHCJS as time goes by. GHCJS is currently more or less compatible with GHC 8.6.x.

I've included GHCJS 8.4 as part of the CI infrastructure in `singletons`, so we should be able to catch any regressions through that.

> Would it be possible to export `PEnum`, `PBounded` from the new minimal `singletons` or that eventual `singlethons-th-lite`?

That seems unlikely, given that we'd have to manually write all of the TH-generated definitions needed to support `PEnum` and `PBounded`, of which there are quite a lot. This would impose a rather large maintenance burden for a minimal `singletons` library, since I would need to keep the manually written definitions up to date with any improvements on the TH generation side (which occur surprisingly frequently).

> I was thinking of simply changing `singlethongs` to embrace the minimal singletons, but if you guys would prefer to keep that code as part of this repo as `singletons-th-lite`, then that's even better.

Honestly, I don't really have a preference, which is why I wanted to ask for your input. `singlethongs` is a relatively small package, so it wouldn't really be that much of a maintenance burden on its own to incorporate its code. My fear is that my design goals might be different than yours (especially _vis-à-vis_ the scope of packages), so perhaps you'd prefer to maintain it yourself?

k0001:
> I've included GHCJS 8.4 as part of the CI infrastructure in `singletons`, so we should be able to catch any regressions through that.

I saw it! This is great! Thank you!

> That seems unlikely, given that we'd have to manually write all of the TH-generated definitions needed to support `PEnum` and `PBounded`, of which there are quite a lot. This would impose a rather large maintenance burden for a minimal `singletons` library, since I would need to keep the manually written definitions up to date with any improvements on the TH generation side (which occur surprisingly frequently).

That makes sense. No worries, I'll see if I can either replicate what I need inside `exinst`, or put this particular functionality in a separate package that depends on the full blown `singletons-base`. 

> Honestly, I don't really have a preference, which is why I wanted to ask for your input. `singlethongs` is a relatively small package, so it wouldn't really be that much of a maintenance burden on its own to incorporate its code. My fear is that my design goals might be different than yours (especially _vis-à-vis_ the scope of packages), so perhaps you'd prefer to maintain it yourself?

There's no need to rush this. I can update `singlethongs` to rely on `singletons` and maintain this code myself. Later on, if more people start using what `singlethongs` offers and we realize that moving that code inside the `singletons` repo would improve things somehow, we can do that then. 

This is all very exciting @RyanGlScott! Thank you for your hard work!

RyanGlScott:
Great, glad to hear that you're on board. I'll keep the package organization the way it currently is, then.

I'd like to run this by @goldfirere before merging.

goldfirere:
I'm agnostic on how, exactly, we achieve this split -- but I'm in full support of splitting, and this particular arrangement seems to satisfy.

My only question is whether the module structure should also be rejiggered. The three packages all have entries in `Data.Singletons.*`. Is that right? And, to be honest, I've never been very happy with putting all the `base` functions in a `Prelude` directory, as they're not really all about the Prelude. (That decision was made a long time ago, and I was focused on functionality over design at the time.) Of course, this rejiggering takes time and -- especially if we rename `Prelude` stuff -- is disruptive to users. So I don't insist. But if others agree about this, then I think now is the time to do it. In this PR or a separate one? I'm not sure -- happy either way.

In any case, I explicitly endorse merging this PR as is.

Thanks!

RyanGlScott:
> The three packages all have entries in `Data.Singletons.*`. Is that right?

That's right. `singletons-th` tries to distinguish itself from `singletons` by using `Data.Singletons.TH.*` for its modules, but then again, it also has exports `Data.Singletons.CustomStar` and `Data.Singletons.SuppressUnusedWarnings`. Additionally, `singletons-base` tries to distinguish itself from `singletons` from by using `Data.Singletons.Prelude*`, but it also has exceptions in `Data.Singletons.TypeError`, `Data.Singletons.TypeLits`, and `Data.Singletons.TypeRepTYPE`.

Thankfully, the exceptions are few. I think we could put everything under a consistent namespace without too much difficulty.

> And, to be honest, I've never been very happy with putting all the `base` functions in a `Prelude` directory, as they're not really all about the `Prelude`.

Indeed, that seems like an especially odd design choice now that we call the package `singletons-base`. I think something like `Data.Singletons.Base.*` would be more appropriate.

> But if others agree about this, then I think now is the time to do it. In this PR or a separate one?

I agree that we should do this before the next major release. This PR is already quite large, so I think I'll merge this now and do the renaming in a separate patch. I've opened #464 to discuss how the new module names should look.

> In any case, I explicitly endorse merging this PR as is.

Hooray!

phadej:
>  I think something like Data.Singletons.Base.*

:+1:

PR #468, opened 2020-06-22
Rename modules in singletons-{th,base}

### LINKED ISSUE #464: More consistent module naming
Now that we've split `singletons` into three packages, it has shined a light on a lurking issue. Namely, the choices of module names. Ideally, it should be easy to tell apart modules from `singletons`, `singletons-th`, and `singletons-base` simply by looking at the module prefix, and to a certain extent, that is true today:

* `singletons` uses `Data.Singletons.*`
* `singletons-th` uses `Data.Singletons.TH.*` for most of its modules
* `singletons-base` uses `Data.Singletons.Prelude.*` for most of its modules

There are two problems, however:

1. The word "most". There are exceptions to the naming rules in `singletons-th` and `singletons-base`:
   * `singletons-th` defines `Data.Singletons.CustomStar` and `Data.Singletons.SuppressUnusedWarnings`, neither of which live in `Data.Singletons.TH.*`.
   * `singletons-base` defines `Data.Singletons.TypeError`, `Data.Singletons.TypeLits`, and `Data.Singletons.TypeRepTYPE`, none of which live in `Data.Singletons.Prelude.*`.
2. `singletons-base` overuses `Data.Singletons.Prelude.*`, especially since many of the things in that module namespace do not correspond to things from the actual `Prelude` (e.g., `Data.Singletons.Prelude.Proxy`).

Given that we're completely reorganizing the package structure in `singletons*-3.0`, now is the time to fix an issue like this. In particular, I propose that we do the following:

1. In `singletons-th`, use `Data.Singletons.TH.*` consistently by renaming `Data.Singletons.CustomStar` and `Data.Singletons.SuppressUnusedWarnings` to `Data.Singletons.TH.CustomStar` and `Data.Singletons.TH.SuppressUnusedWarnings`, respectively.
2. In `singletons-base`, use the module namespace `Data.Singletons.Base.*` to make it mesh better with the name of the library. In particular, I propose to rename most modules from `Data.Singletons.Prelude.X` to `Data.Singletons.Base.X` _carte blanche_. An exception to this rule would be `Data.Singletons.Prelude` itself, which I would propose renaming to `Data.Singletons.Base.Prelude`.

  I would also propose renaming `Data.Singletons.TypeError`, `Data.Singletons.TypeLits`, and `Data.Singletons.TypeRepTYPE` to `Data.Singletons.Base.TypeError`, `Data.Singletons.Base.TypeLits`, and `Data.Singletons.Base.TypeRepTYPE`, respectively.

What do others think of this plan?

RyanGlScott:
I've split this up into two commits to make this slightly easier to review.

Fixes #464.


README.md
```diff
@@ -118,26 +118,26 @@ equality. See the "Equality classes" section for more information.
 
 `Data.Singletons.TH` (from `singletons-th`) exports all the definitions needed
 to use the Template Haskell code to generate new singletons.
-`Data.Singletons.Prelude.TH` (from `singletons-bsae`) re-exports
+`Data.Singletons.Base.TH` (from `singletons-bsae`) re-exports
```
goldfirere, line 121:
```suggestion
`Data.Singletons.Base.TH` (from `singletons-base`) re-exports
```

singletons-base/CHANGES.md
```diff
@@ -13,24 +13,56 @@ Changelog for singletons-base project
...
   specific to those libraries. For more information on this split, see the
   [relevant GitHub discussion](https://github.com/goldfirere/singletons/issues/420).
 * Require building with GHC 8.12.
-* Due to the `singletons` package split, the `Data.Singletons.CustomStar` and
-  `Data.Singletons.TH` modules in `singletons-th` no longer re-export any
-  definitions from `Data.Singletons.Prelude.*`. The `singletons-base` library
-  now provides versions of these modules—`Data.Singletons.Prelude.CustomStar`
-  and `Data.Singletons.Prelude.TH`, respectively—that do re-export
-  definitions from `Data.Singletons.Prelude.*`.
+* The modules in `singletons-base` have been renamed to better reflect the
```
goldfirere, line 23: I don't see a mention anywhere about the highly controversial `Data.Singletons.Base.Enum` module. Perhaps add it?

README.md
RyanGlScott: Eep, good catch. Thankfully, this appears to be the only occurrence of this typo (according to `grep`).

singletons-base/CHANGES.md
RyanGlScott: I've added a bit below here that explains this a bit more.

PR #471, opened 2020-06-25
Preserve strict data type fields when singling

### LINKED ISSUE #470: singletons-th discards strictness in data type fields
Consider this example:

```hs
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE GADTs #-}
{-# LANGUAGE PolyKinds #-}
{-# LANGUAGE StandaloneKindSignatures #-}
{-# LANGUAGE TypeApplications #-}
{-# LANGUAGE TypeFamilies #-}
{-# OPTIONS_GHC -Wall #-}
module Bug where

import Data.Kind
import Data.Singletons.TH

$(singletons [d|
  type T :: Type -> Type
  data T a where
    MkT1 :: a -> T a
    MkT2 :: !Void -> T a
  |])

f :: T a -> a
f (MkT1 x) = x

type F :: T a -> a
type family F x where
  F ('MkT1 x) = x

sF :: forall a (x :: T a).
      Sing x -> Sing (F x)
sF (SMkT1 sx) = sx
```

When compiled, `f` will emit no warnings. But `sF`, its singled counterpart, will emit a warning:

```
Bug.hs:32:1: warning: [-Wincomplete-patterns]
    Pattern match(es) are non-exhaustive
    In an equation for ‘sF’: Patterns not matched: SMkT2 _
   |
32 | sF (SMkT1 sx) = sx
   | ^^^^^^^^^^^^^^^^^^
```

Why the discrepancy? It's because of one key different between the declarations for `T` and its singled counterpart, `ST`. The `MkT2` constructor of `T` declares its field of type `Void` as strict, which convinces GHC's pattern-match coverage checker that one cannot match on `MkT2` in `f`. On the hand, `ST` is declared like so:

```hs
type ST :: forall (a :: Type). T (a :: Type) -> Type
data ST z
  where
    SMkT1 :: forall a (n :: a). (Sing n) -> ST (MkT1 n :: T a)
    SMkT2 :: forall a (n :: Void). (Sing n) -> ST (MkT2 n :: T a)
```

This does _not_ declare `SMkT2`'s field as strict, which means that `sF` really is non-exhaustive, since one could invoke `sF (SMkT2 ⊥)`. I don't see any reason we shouldn't just carry strictness annotation from `MkT2` over to `SMkT2`, which would fix this issue.

RyanGlScott:
Fixes #470.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1006,6 +1006,21 @@ instance C [] where
   method = []
 ```
 
+## Support for singling, but not promotion
+
+The following constructs are supported for singleton generation but not promotion:
+
+* bang patterns
+
+See the following sections for more details.
+
+### Bang patterns
+
+Bang patterns (e.g., `f !x = (x, x)`) cannot be translated to a type-level
+setting as type families lack an equivalent of bang patterns. As a result,
+`singletons-th` will ignore any bang patterns and will simply promote the
+underyling pattern instead.
+
 ## Little to no support
 
 The following constructs are either unsupported or almost never work:
@@ -1015,6 +1030,8 @@ The following constructs are either unsupported or almost never work:
 * rank-n types
 * promoting `TypeRep`s
 * `TypeApplications`
+* Irrefutable patterns
+* `{-# UNPACK #-}` pragmas
 
 See the following sections for more details.
 
@@ -1322,3 +1339,25 @@ for the following constructs:
   type MSym0 :: forall a b. a ~> b ~> a
   type MSym1 :: forall a b. a -> b ~> a
   ```
+
+### Irrefutable patterns
+
+`singletons-th` will ignore irrefutable patterns (e.g., `f ~(x, y) = (y, x)`)
+and will simply promote or single the underlying patterns instead.
+`singletons-th` cannot promote irrefutable patterns for the same reason it
+cannot promote bang patterns: there is no equivalent syntax for type families.
+Moreover, `singletons-th` cannot single irrefutable patterns since singled data
+constructors are implemented as GADTs, as irrefutably matching on a GADT
+constructor will not bring the underlying type information into scope. Since
+essentially all singled code relies on using GADT type information in this way,
+it cannot reasonably be combined with irrefutable patterns, which prevent this
+key feature of GADT pattern matching.
+
+### `{-# UNPACK #-}` pragmas
+
+`singletons-th` will ignore `{-# UNPACK #-}` pragmas on the fields of a data
+constructor (e.g., `data T = MkT {-# UNPACK #-} !()`). This is because
+singled data types represent their argument types using existential type
+variables, and any data constructor that explicitly uses existential
+quantification cannot be unpacked. See
+[GHC#10016](https://gitlab.haskell.org/ghc/ghc/-/issues/10016).
--- a/singletons-base/tests/SingletonsBaseTestSuite.hs
+++ b/singletons-base/tests/SingletonsBaseTestSuite.hs
@@ -137,6 +137,7 @@ tests =
     , compileAndDumpStdTest "T450"
     , compileAndDumpStdTest "T453"
     , compileAndDumpStdTest "NegativeLiterals"
+    , compileAndDumpStdTest "T470"
     ],
     testCompileAndDumpGroup "Promote"
     [ compileAndDumpStdTest "Constructors"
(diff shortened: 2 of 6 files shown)


RyanGlScott:
Hm. I was under the impression that `{-# UNPACK #-}` didn't work on _any_ GADTs when submitting this, but now I realize that that's not quite true. Here is a counterexample to that claim:

```hs
data SFoo = SMkFoo {-# UNPACK #-} !(STuple0 '())
```

If you look at the resulting Core in `-ddump-simpl`, this actually does unpack:

```
-- RHS size: {terms: 5, types: 9, coercions: 1, joins: 0/0}
Bug.$WSMkFoo [InlPrag=INLINE[0]] :: STuple0 '() -> SFoo
[GblId[DataConWrapper],
 Arity=1,
 Caf=NoCafRefs,
 Str=<S,U>m,
 Unf=Unf{Src=InlineStable, TopLvl=True, Value=True, ConLike=True,
         WorkFree=True, Expandable=True,
         Guidance=ALWAYS_IF(arity=1,unsat_ok=True,boring_ok=False)
         Tmpl= \ (dt_aGp [Occ=Once!] :: STuple0 '()) ->
                 case dt_aGp of { STuple0 dt_aGq ->
                 Bug.SMkFoo @~ (<'()>_N :: '() GHC.Prim.~# '())
                 }}]
Bug.$WSMkFoo
  = \ (dt_aGp [Occ=Once!] :: STuple0 '()) ->
      case dt_aGp of { STuple0 dt_aGq ->
      Bug.SMkFoo @~ (<'()>_N :: '() GHC.Prim.~# '())
      }
```

However, this doesn't always hold. For instance, this does _not_ unpack:

```hs
data SBar = SMkBar {-# UNPACK #-} !(SIdentity ('Identity '()))
```
```
warning:
    • Ignoring unusable UNPACK pragma on the first argument of ‘SMkBar’
    • In the definition of data constructor ‘SMkBar’
      In the data type declaration for ‘SBar’
   |
   | data SBar = SMkBar {-# UNPACK #-} !(SIdentity ('Identity '()))
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Perhaps I should just let `{-# UNPACK #-}` pragmas through wholesale, [GHC#10016](https://gitlab.haskell.org/ghc/ghc/-/issues/10016) notwithstanding. After all, GHC will warn if it ultimately fails to unpack.

RyanGlScott:
Please disregard what I wrote in https://github.com/goldfirere/singletons/pull/471#issuecomment-649503207; it's not really applicable to `singletons-th`. Indeed, `singletons-th` will never generate definitions like `SFoo` and `SBar` above. In fact, any time it singles a data type with at least one argument, it will create an existential type variable. And GHC#10016 triggers precisely when there is at least one existential type variable in the wrapper, which is why `SFoo` above unpacks (since it only uses an existential type variable in the _worker_, not the wrapper).

README.md
```diff
@@ -1015,6 +1015,7 @@ The following constructs are either unsupported or almost never work:
 * rank-n types
 * promoting `TypeRep`s
 * `TypeApplications`
+* `{-# UNPACK #-}` pragmas
```
goldfirere, line 1018: Can't we just drop these pragmas? That seems strictly more expressive, allowing users to unpack their unrefined datatypes, even if we don't do this for the singletons we generate.

EDIT: Actually, it seems this is what we do, from reading the code. But then this README is misleading.

RyanGlScott: The treatment of `UNPACK` patterns here is basically identical to how we treat bang patterns during promotion—we drop them and emit a warning saying as much. (Granted, the fact that we drop bang patterns during promotion is not documented in the `README`. It should be.)

How would you phrase this to be less misleading?

goldfirere: It's this text: 

```
Attempting to single a data constructor with an `{-# UNPACK #-}` pragma
(e.g., `data T = MkT {-# UNPACK #-} !()`) will not work.
```

Maybe just say the `{-# UNPACK #-}` will be ignored.

RyanGlScott: Ah. I think I intended that to say "won't work as you might expect", but it ended up sounding more uncompromising than I had intended. I've reworded this to use your suggesting phrasing, and I've adopted similar phrasing in the newly minted sections on bang patterns and irrefutable patterns.

PR #474, opened 2020-09-29
Update th-desugar submodule to ca76128

RyanGlScott:
This updates the `th-desugar` submodule to commit goldfirere/th-desugar@ca761285741c16c704a4044f551faf4292432d6a, which incorporates several recent developments:

* `DTyVarBndr`s are now parameterized by a `flag` type parameter. This has several ripple effects throughout `singletons`, as various functions now need to care about `Specificity`.
* `decToTH` returns a `Dec` rather than `[Dec]`. This is used in a handful of places in `singletons` and need to be updated accordingly.
* The type of `dsTvb` is different on pre-9.0 versions of GHC, so we replace uses of `dsTvb` with `dsTvbUnit` in `singletons-th` to ensure backwards compatibility. When we drop support for pre-9.0 versions of GHC in `singletons-th`, we can revisit this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@
 #
 # For more information, see https://github.com/haskell-CI/haskell-ci
 #
-# version: 0.10.1
+# version: 0.10.3
 #
 version: ~> 1.0
 language: c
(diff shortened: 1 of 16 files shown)


RyanGlScott:
CI is failing because `th-desugar` now requires `th-abstraction >= 0.4`, but some `singletons-base-test-suite` dependencies still require `th-abstraction < 0.4`. I'll park this PR until I've sorted out the dependency issues downstream.

goldfirere:
A quick scroll-through shows nothing strange. Let me know if you want particular attention anywhere.

Issue #439, opened 2020-02-19
GHC 9.0 checklist


RyanGlScott:
This issue serves as a reminder to do certain things before we ship a new major release that supports GHC 9.0:

* [x] Update `th-desugar` (see goldfirere/th-desugar#141)
* [x] Consider changing the definition of `ShowSing` to make it easier to derive (#426)
* [x] Make `OptionsM` an instance of the new `Quote` class from `Language.Haskell.TH.Syntax`. That way, we can omit the use of `lift` in code like this from the `README`:

   ```hs
   import Control.Monad.Trans.Class
   import Data.Singletons.TH
   import Data.Singletons.TH.Options
   import Language.Haskell.TH (Name, mkName, nameBase)
   
   $(let myPrefix :: Name -> Name
         myPrefix name = mkName ("MyS" ++ nameBase name) in
   
         withOptions defaultOptions{singledDataConName = myPrefix} $
         singletons $ lift [d| data T = MkT |])
   ```
* [X] Simplify some code using [explicit specificity](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0099-explicit-specificity.rst):

  * The type of `demote`:

    https://github.com/goldfirere/singletons/blob/efb2e395353636f07b95c908792b5eeac6e79d37/src/Data/Singletons/Internal.hs#L602-L603

    Can be made cleaner:

    ```hs
    demote :: forall {k} (a :: k). (SingKind k, SingI a) => Demote k
    ```
  * `SingI`:

    https://github.com/goldfirere/singletons/blob/efb2e395353636f07b95c908792b5eeac6e79d37/src/Data/Singletons/Internal.hs#L115-L118

    Can be given a standalone kind signature:

    ```hs
    type  SingI :: forall {k}. k -> Constraint
    class SingI a where ...
    ```
* [x] `Note [Preserve the order of type variables during singling]` describes several ugly hacks that are required due to Template Haskell's inability to distinguish specified from inferred type variables. Here is one such hack, described in `Wrinkle 2: The TH reification swamp`:

    https://github.com/goldfirere/singletons/blob/f6c4223d3674ea323346ecb03b19c389bac2cdc8/src/Data/Singletons/Single/Type.hs#L216-L220

    Here is another, described in `Wrinkle 3: Where to put explicit kind annotations`:

[...]

PR #483, opened 2021-02-19
Require GHC 9.0 in singletons-{th,base}

RyanGlScott:
This is in preparation for some changes that require GHC 9.0 as the minimum GHC version. See #439.


singletons-base/singletons-base.cabal
```diff
@@ -1,14 +1,14 @@
...
+cabal-version:  1.24
 synopsis:       A promoted and singled version of the base library
 homepage:       http://www.github.com/goldfirere/singletons
 category:       Dependent Types
 author:         Richard Eisenberg <rae@cs.brynmawr.edu>, Jan Stolarek <jan.stolarek@p.lodz.pl>
 maintainer:     Ryan Scott <ryan.gl.scott@gmail.com>
 bug-reports:    https://github.com/goldfirere/singletons/issues
 stability:      experimental
-tested-with:    GHC == 8.10.1
+tested-with:    GHC == 8.10.3
```
int-index, line 11: 9.0?

RyanGlScott: Morally, this ought to say `tested-with: GHC == 9.0.1`. Unfortunately, I'm not able to do this in practice yet, since `haskell-ci` hasn't caught up to 9.0.1 yet—see my comments in `cabal.haskell-ci`. Until then, I'm temporarily hacking around this restriction with `haskell-ci`'s `github-patches` feature, with the intention that I'll remove this hack once haskell-CI/haskell-ci#474 is resolved. (That issue is about `haskell-ci` lacking 8.10.4 support, but it applies equally to 9.0.1 as well.)

int-index: I don’t get it. What would go wrong if you simply changed it to `GHC == 9.0.1` by hand?

RyanGlScott: `haskell-ci` uses `tested-with` information to determine which versions of GHC to use when generating a CI configuration file. Moreover, it's very picky about which versions of GHC you can use. If you use a version of GHC that it doesn't know about yet, such as GHC 9.0.1, then it will simply error out when you try to run `haskell-ci regenerate`:

```
$ haskell-ci regenerate
No haskell-ci.sh, skipping bash regeneration
*ERROR* 'tested-with:' specifically refers to unknown 'GHC' versions: 9.0.1
Known GHC versions: 7.0.1, 7.0.2, 7.0.3, 7.0.4, 7.2.1, 7.2.2, 7.4.1, 7.4.2, 7.6.1, 7.6.2, 7.6.3, 7.8.1, 7.8.2, 7.8.3, 7.8.4, 7.10.1, 7.10.2, 7.10.3, 8.0.1, 8.0.2, 8.2.1, 8.2.2, 8.4.1, 8.4.2, 8.4.3, 8.4.4, 8.6.1, 8.6.2, 8.6.3, 8.6.4, 8.6.5, 8.8.1, 8.8.2, 8.8.3, 8.8.4, 8.10.1, 8.10.2, 8.10.3
```

I'd prefer not to have make manual changes each time I update the CI configuration, so I decided to just hack my way around this for now, especially since there are other upcoming changes that depend on this one. I expect this hack to be short-lived, as we can remove it once `haskell-ci` supports GHC 9.0.1 properly.

RyanGlScott: haskell-CI/haskell-ci#474 is now fixed, so I have pushed an update that properly specifies `tested-with: GHC == 9.0.1`.

PR #486, opened 2021-02-22
Tweak ShowSing' so that Show instances for singletons can be derived

### LINKED ISSUE #426: Consider an alternative version of ShowSing that can be derived more easily
When [`Sing` became a type family](https://github.com/goldfirere/singletons/issues/318), we had to surgically alter the `ShowSing` class to make it possible to mention `Sing` in the head of a quantified constraint:

```hs
class    (forall (z :: k). ShowSing' z) => ShowSing k
instance (forall (z :: k). ShowSing' z) => ShowSing k

class    Show (Sing z) => ShowSing' z
instance Show (Sing z) => ShowSing' z
```

This came at a cost, however. It is no longer possible to derive `Show` instances for `Sing` types like this:

```hs
deriving instance ShowSing k => Show (SBlah (z :: k))
```

Instead, one has to write out the full implementation of `showsPrec` like so in order to make it typecheck:

```hs
instance ShowSing k => Show (SBlah (z :: k)) where
  showsPrec p (SMkBlah (s :: Sing k)) =
    showParen (p > 10) $ showString "SMkBlah " . showsPrec 11 s
      :: ShowSing' k => ShowS
```

See [here](https://github.com/goldfirere/singletons/blob/013264115564b707f7cf440e97e9a3387dfb9987/src/Data/Singletons/ShowSing.hs#L116-L194) for an explanation. This is not the end of the world, since most `Show` instances for `Sing`s are autogenerated through Template Haskell anyway, but it does make implementing the few hand-written `Show` instances for `Sing`s much more annoying.

What if we could have our cake and eat it too? That is, what if we were able to tweak `ShowSing` so that it _could_ be used in standalone `deriving` declarations like the example I pose above? This may in fact be possible once https://gitlab.haskell.org/ghc/ghc/merge_requests/2283 lands in GHC HEAD, which tweaks how quantified constraints work to make it feasible for `singletons` to use the following version of `ShowSing`:

```hs
class    (forall (z :: k). ShowSing' z) => ShowSing k
instance (forall (z :: k). ShowSing' z) => ShowSing k

class    (forall (sing :: k -> Type). sing ~ Sing => Show (sing z))
                       => ShowSing' (z :: k)
instance Show (Sing z) => ShowSing' (z :: k)
```

As the patch for https://gitlab.haskell.org/ghc/ghc/merge_requests/2283 demonstrates, this allows the following code to typecheck:

```hs
newtype Foo a = MkFoo a
data SFoo :: forall a. Foo a -> Type where
  SMkFoo :: Sing x -> SFoo (MkFoo x)
type instance Sing = SFoo
deriving instance ShowSing a => Show (SFoo (z :: Foo a))

newtype Bar a = MkBar (Foo a)
data SBar :: forall a. Bar a -> Type where
  SMkBar :: forall a (x :: Foo a). Sing x -> SBar (MkBar x)
type instance Sing = SBar
deriving instance ShowSing (Foo a) => Show (SBar (z :: Bar a))
```

This is very exciting, as there is a very real possibility that this could work for everything in `singletons`. I decided to try this out myself to confirm this hunch.

Did it work? Yes! Well, mostly. There ~~were a couple of hiccups~~ was one hiccup along the way:

1. This instance:

   https://github.com/goldfirere/singletons/blob/013264115564b707f7cf440e97e9a3387dfb9987/src/Data/Singletons/ShowSing.hs#L274-L277

   No longer typechecks, even when I rewrite it to `deriving instance ShowSing k => Show (WrappedSing (a :: k))`:

    ```
   src/Data/Singletons/ShowSing.hs:277:1: error:
       • Couldn't match type ‘Sing’ with ‘WrappedSing’
           arising from a use of ‘GHC.Show.$dmshow’
       • In the expression: GHC.Show.$dmshow @(WrappedSing a)
         In an equation for ‘show’: show = GHC.Show.$dmshow @(WrappedSing a)
         When typechecking the code for ‘show’
           in a derived instance for ‘Show (WrappedSing a)’:
           To see the code I am typechecking, use -ddump-deriv
         In the instance declaration for ‘Show (WrappedSing a)’
       • Relevant bindings include
           show :: WrappedSing a -> String
             (bound at src/Data/Singletons/ShowSing.hs:277:1)
       |
   277 | deriving instance ShowSing k => Show (WrappedSing (a :: k))
       | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   
   src/Data/Singletons/ShowSing.hs:277:1: error:
       • Couldn't match type ‘Sing’ with ‘WrappedSing’
           arising from a use of ‘GHC.Show.$dmshowList’
       • In the expression: GHC.Show.$dmshowList @(WrappedSing a)
         In an equation for ‘showList’:
             showList = GHC.Show.$dmshowList @(WrappedSing a)
         When typechecking the code for ‘showList’
           in a derived instance for ‘Show (WrappedSing a)’:
           To see the code I am typechecking, use -ddump-deriv
         In the instance declaration for ‘Show (WrappedSing a)’
       • Relevant bindings include
           showList :: [WrappedSing a] -> ShowS
             (bound at src/Data/Singletons/ShowSing.hs:277:1)
       |
   277 | deriving instance ShowSing k => Show (WrappedSing (a :: k))
       | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   ```

   Very strange. I was able to make this typecheck again through the very painful process of inlining the default implementations of `show` and `showList`:

   ```hs
   instance ShowSing k => Show (WrappedSing (a :: k)) where
     showsPrec p (WrapSing s) = showParen (p >= 11) $ showsSing s
   
     show (WrapSing s) = showsSing s ""
   
     showList []              s = "[]" ++ s
     showList (WrapSing x:xs) s = '[' : showsSing x (showl xs)
       where
         showl []              = ']' : s
         showl (WrapSing y:ys) = showsSing y (showl ys)
   
   showsSing :: ShowSing k => Sing (x :: k) -> ShowS
   showsSing s = showString "WrapSing {unwrapSing = " . showsPrec 0 s . showChar '}'
   ```

   It feels very odd having to do this, however. I'm unclear if this is a GHC bug or a fundamental limitation of how quantified constraints work. (Note that like `Sing`, `WrappedSing` is also of kind `k -> Type`, which makes me wonder if that causes the local `(forall (sing :: k -> Type). sing ~ Sing => Show (sing z))` constraint to be fired too early.)
2. ~~I couldn't figure out how to make this instance typecheck at all, even with modifications:~~

   https://github.com/goldfirere/singletons/blob/013264115564b707f7cf440e97e9a3387dfb9987/src/Data/Singletons/Sigma.hs#L80-L86

   **EDIT**: Never mind, I was being silly here. I was trying to be too clever by changing `ShowSingApply'` to be more like `ShowSing'`.

Everything else Just Worked™. Still, the ~~two drawbacks~~ one drawback listed above makes me wonder if this idea is completely baked yet, so I opted to simply open an issue for discussing the idea before moving forward with a patch.

RyanGlScott:
By giving `ShowSing'` a quantified superclass constraint, it becomes possible to derive `Show` instances for singleton types that mention `ShowSing` in the instance context. This is a remarkably backwards-compatible change except for one corner case, which I have documented in the Haddocks for `ShowSing'` in `Data.Singletons.ShowSing`.

Fixes #426, as well as one bullet point of #439.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,7 @@ includes, but is not necessarily limited to, the following:
 * `PolyKinds`
 * `RankNTypes`
 * `ScopedTypeVariables`
+* `StandaloneDeriving`
 * `StandaloneKindSignatures`
 * `TemplateHaskell`
 * `TypeApplications`
--- a/singletons-base/src/Data/Singletons/Base/Instances.hs
+++ b/singletons-base/src/Data/Singletons/Base/Instances.hs
@@ -11,7 +11,7 @@ re-exported from various places.
 {-# LANGUAGE DataKinds, PolyKinds, RankNTypes, GADTs, TypeFamilies, EmptyCase,
              FlexibleContexts, TemplateHaskell, ScopedTypeVariables,
              UndecidableInstances, TypeOperators, FlexibleInstances,
-             TypeApplications, StandaloneKindSignatures #-}
+             TypeApplications, StandaloneKindSignatures, StandaloneDeriving #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 
 module Data.Singletons.Base.Instances (
--- a/singletons-base/src/Data/Singletons/Base/SomeSing.hs
+++ b/singletons-base/src/Data/Singletons/Base/SomeSing.hs
@@ -2,6 +2,7 @@
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE PolyKinds #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 
 -----------------------------------------------------------------------------
@@ -73,10 +74,7 @@ instance SNum k => Num (SomeSing k) where
   signum (SomeSing a) = SomeSing (sSignum a)
   fromInteger n = withSomeSing (fromIntegral n) (SomeSing . sFromInteger)
 
-instance ShowSing k => Show (SomeSing k) where
-  showsPrec p (SomeSing (s :: Sing a)) =
-    showParen (p > 10) $ showString "SomeSing " . showsPrec 11 s
-      :: ShowSing' a => ShowS
+deriving instance ShowSing k => Show (SomeSing k)
 
 instance SSemigroup k => Semigroup (SomeSing k) where
   SomeSing a <> SomeSing b = SomeSing (a %<> b)
--- a/singletons-base/tests/compile-and-dump/GradingClient/Database.golden
+++ b/singletons-base/tests/compile-and-dump/GradingClient/Database.golden
@@ -3357,51 +3357,10 @@ GradingClient/Database.hs:(0,0)-(0,0): Splicing declarations
                                                         -> Type) where
       Data.Type.Coercion.testCoercion
         = Data.Singletons.Decide.decideCoercion
-    instance (Data.Singletons.ShowSing.ShowSing U,
-              Data.Singletons.ShowSing.ShowSing Nat) =>
-             Show (SU (z :: U)) where
-      showsPrec _ SBOOL = showString "SBOOL"
-      showsPrec _ SSTRING = showString "SSTRING"
-      showsPrec _ SNAT = showString "SNAT"
-      showsPrec
-        p_0123456789876543210
-        (SVEC (arg_0123456789876543210 :: Sing argTy_0123456789876543210)
-              (arg_0123456789876543210 :: Sing argTy_0123456789876543210))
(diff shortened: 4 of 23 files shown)


goldfirere (approved):
This change is a bit gross, but if it works in practice, it's perhaps no worse than other unsavory tricks of `singletons`. We really need to fix https://gitlab.haskell.org/ghc/ghc/-/issues/16365#note_189057, but let's not hold this up on that fix.

PR #494, opened 2021-04-10
Add Data.Functor.{Compose,Product,Sum}.Singletons

### LINKED ISSUE #491: Add missing promotions
There are some things we can promote that we can't singletonize. For example, I don't see any obvious problem with

```haskell
$(genPromotions [''Compose, ''Product])
$(promoteOnly [d|
  instance (Functor f, Functor g) => Functor (Compose f g) where
    fmap fun (Compose x) = Compose (fmap (fmap fun) x)

  instance (Functor f, Functor g) => Functor (Product f g) where
    fmap fun (Pair x y) = Pair (fmap fun x) (fmap fun y)
  |])
```

even though `genSingletons` fails for both `Compose` and `Product`.

I believe we should add such, just for folks using `singletons` facilities for type-level programming (e.g., the automatic defunctionalization).

RyanGlScott:
Fixes #491.

### DIFF
--- a/singletons-base/CHANGES.md
+++ b/singletons-base/CHANGES.md
@@ -3,6 +3,7 @@ Changelog for the `singletons-base` project
 
 next [????.??.??]
 -----------------
+* Add `Data.Functor.{Compose,Product,Sum}.Singletons`.
 * The types of various entities in `Data.Functor.Const.Singletons` and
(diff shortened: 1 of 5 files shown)


treeowl:
Your knowledge here is way broader and deeper than mine. If you think it's good, it's probably good.

PR #501, opened 2021-07-05
Add `SingI1` and `SingI2` classes

### LINKED ISSUE #500: Introduce `SingI1`, a second-order version of `SingI`
We were replacing `Typeable` with`SingI` in our project. We realised that having `SingI` is not enough in several places since we also need `SingI` for type constructors. So we introduced the class `SingI1` for lifting `SingI` to unary constructors in the same fashion as, for example, `Eq1` or `Ord1`.

```haskell
class SingI1 f where
  withSingI1 :: forall x r. SingI x => (SingI (f x) => r) -> r
```

The idea is to introduce this type class in the library.

RyanGlScott:
This adds `SingI{1,2}`, higher-order versions of the `SingI` class, to `Data.Singletons`. This has a couple of knock-on effects:

* `Data.Singletons` also sports a variety of derived functions which make it more convenient to use `SingI{1,2}`. See the `singletons` changelog for more details.
* `singletons-th` now generates `SingI{1,2}` instances whenever possible in addition to generating `SingI` instances. This required some mild refactoring to accomplish, but nothing too major.

Fixes #500.

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,6 +184,7 @@ The type family of singleton types. A new instance of this type family is
 generated for every new singleton type.
 
 ```haskell
+type SingI :: forall {k}. k -> Constraint
 class SingI a where
   sing :: Sing a
 ```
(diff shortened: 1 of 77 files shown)


DanielRrr (approved):
Looks good to me. I'm really not an expect in singletons though. Perhaps someone else might have remarks or smth like that.

DanielRrr:
cc @int-index ^

RyanGlScott:
If this works for you, then I'm quite happy to merge this.

Issue #487, opened 2021-03-14
Use type-level characters in Text.Show.Singletons


RyanGlScott:
GHC 9.2 will feature [type-level `Char`s](https://gitlab.haskell.org/ghc/ghc/-/commit/7f3524efcbd58ca6837ec0ffca6ddd121d64e4de). This will likely prove useful for `singletons-base`'s purposes, as the `Text.Show.Singletons` module currently employs several nasty hacks to work around the lack of type-level `Char`s. This code in particular speaks volumes:

https://github.com/goldfirere/singletons/blob/38eac92d311ebd188d008190c84a20b664e2ab9d/singletons-base/src/Text/Show/Singletons.hs#L81-L84

I've started a [`type-level-chars` branch](https://github.com/goldfirere/singletons/tree/type-level-chars) which explores refactoring `singletons-base` to remove the `SChar` hack and just using the `Char` kind. I was able to do quite a bit with the current state of GHC HEAD, but I wasn't able to promote quite everything in `Text.Show` yet. In particular, I'm not able to promote the `Show Char` instance due to not having a type-level `ord` function. Such a function will be added in [this MR](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3598), but it's not yet clear to me if that will land in time for GHC 9.2.1.

When GHC 9.2.1 is released, we should revisit this and see how many hacks can be removed from `Text.Show.Singletons`.

int-index:
@RyanGlScott Do you need anything besides `ord` to make this work? One of the issues with the MR you linked to is that it adds a lot of built-in type families, and it’s not clear whether we need them all.

But `ord` should be easy (and `chr` for symmetry, I suppose). If that’s all you need, we could easily extract them into a smaller MR.

RyanGlScott:
> Do you need anything besides `ord` to make this work?

I'm not sure, as I didn't explore any further once I ran into the `ord` obstacle. But perhaps I should try building a stripped-down version of [this MR](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3598) that only defines type-level `ord` and `chr` and see if those are enough to remove all of the hacks currently used in `Text.Show.Singletons`. (I'm loosely defining "hack" to mean "using `Symbol` in a place where `Char` ought to be used instead".)

int-index:
Alright, I’ll prepare a branch for this, and we can add more type families to it if you discover a need for them.

int-index:
[...]

PR #506, opened 2021-09-25
Add more documentation for overlapping and literal patterns

RyanGlScott:
While working on #487, I realized that the current `README` documentation about literal patterns is non-existent, so I decided to add some. In order to explain the limitations of literal patterns, however, one must explain the limitations of overlapping patterns, and I discovered that the `README` also doesn't do a very good job of explaining this either. At this point, I decided to spin off this documentation improvement into its own commit.


goldfirere (approved):
Looks great -- nice contribution.

Thanks!

PR #507, opened 2021-09-25
Support type-level `Char`s

### LINKED ISSUE #487: Use type-level characters in Text.Show.Singletons

RyanGlScott:
GHC 9.2 now supports type-level characters, which means that `singletons-th` is now able to promote and single character literals in certain circumstances. It also means that we can remove two terrible hacks in `Text.Show.Singletons`:

* We no longer need `SChar`, which fakes type-level characters with `Symbol`s of length one.
* The `PShow` and `SShow` instances for `Symbol` can now display escape characters properly rather than returning the input `Symbol` unchanged.

Fixes #487.

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,13 +305,6 @@ generate these instances directly through functions exported from
 `Data.Singletons.TH` (from `singletons-th`) and
 `Data.Singletons.Base.TH` (from `singletons-base`).
 
-A promoted and singled `Show` instance is provided for `Symbol`, but it is only
-a crude approximation of the value-level `Show` instance for `String`. On the
-value level, showing `String`s escapes special characters (such as double
-quotes), but implementing this requires pattern-matching on character literals,
-something which is currently impossible at the type level. As a consequence, the
-type-level `Show` instance for `Symbol`s does not do any character escaping.
-
 Errors
 ------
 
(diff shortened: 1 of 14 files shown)


goldfirere (approved):
Looks reasonable to me. I haven't thought about what other definitions, etc., users might want (maybe more stuff from `Data.Char`?), but this is at least a good start. I recommend merging even if we might need to add more later.

RyanGlScott:
> I haven't thought about what other definitions, etc., users might want (maybe more stuff from `Data.Char`?), but this is at least a good start.

Yeah, it would be neat to have `Data.Char.Singletons` at some point. As things currently stand, we wouldn't be able to offer promoted versions of most of `Data.Char`, since we would need the functionality from [GHC!3598](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3598) to do so. If that lands, however, it would be worth revisiting this idea.

PR #514, opened 2022-01-27
Provide more specific error message for partial `(->)` applications

### LINKED ISSUE #511: Poor error message with partial application of function arrow
If I write this:

```hs
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE GADTs #-}
{-# LANGUAGE PolyKinds #-}
{-# LANGUAGE StandaloneKindSignatures #-}
{-# LANGUAGE TypeApplications #-}
{-# LANGUAGE TypeFamilies #-}
module Bug where

import Data.Proxy
import Data.Proxy.Singletons
import Data.Singletons.TH

$(singletons [d|
  data Foo = MkFoo (Proxy (->))
  |])
```

Then `singletons-th` will produce this error:

```
[1 of 1] Compiling Bug              ( Dyn.hs, interpreted )

Dyn.hs:15:2: error: Q monad failure
   |
15 | $(singletons [d|
   |  ^^^^^^^^^^^^^^^...

Dyn.hs:15:2: error:
    Illegal Haskell construct encountered:
        headed by: DArrowT
        applied to: []

   |
15 | $(singletons [d|
   |  ^^^^^^^^^^^^^^^...
```

This is a poor error message, as there's certainly nothing illegal about the program we're trying to promote. It may be the case that `singletons-th` can't successfully handle something like this, but if so, the error message should just state that.

This was originally spotted in #510, but this issue is separate from the main problem being discussed in #510.

RyanGlScott:
Fixes #511.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1277,6 +1277,7 @@ The following constructs are either unsupported or almost never work:
 * `TypeApplications`
 * Irrefutable patterns
 * `{-# UNPACK #-}` pragmas
+* partial application of the `(->)` type
 
 See the following sections for more details.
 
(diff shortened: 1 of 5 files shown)


goldfirere:
LGTM. Thanks!

PR #522, opened 2022-05-14
Use `withDict` to implement `withSingI` on recent GHCs

### LINKED ISSUE #490: Why singletons use unsafeCoerce and not magicDict
See https://gitlab.haskell.org/ghc/ghc/-/blob/master/compiler/GHC/Types/Id/Make.hs#L1804-1844

Would it work?

RyanGlScott:
Instead of `unsafeCoerce`, we can use the more recent `withDict` primitive from `GHC.Exts`. Among other benefits, this is more likely to be future-proof against possible changes to GHC's representation of "newtype classes" in the future.

Fixes #490.

### DIFF
--- a/singletons/src/Data/Singletons.hs
+++ b/singletons/src/Data/Singletons.hs
@@ -121,6 +121,10 @@ import Data.Proxy (Proxy(..))
 import GHC.Exts (Proxy#)
 import Unsafe.Coerce (unsafeCoerce)
 
+#if MIN_VERSION_base(4,17,0)
+import GHC.Exts (withDict)
+#endif
+
 -- | Convenient synonym to refer to the kind of a type variable:
 -- @type KindOf (a :: k) = k@
 #if __GLASGOW_HASKELL__ >= 810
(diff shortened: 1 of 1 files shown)


RyanGlScott:
I decided to try and tackle `reflection`'s uses of `unsafeCoerce` after this, at which point I discovered that replacing `unsafeCoerce` with `withDict` broke `reflection`'s test suite. See [GHC#21575](https://gitlab.haskell.org/ghc/ghc/-/issues/21575). As a result, I'm going to mark this PR as a draft and wait for a resolution to that issue before proceeding.

RyanGlScott:
While [GHC#21575](https://gitlab.haskell.org/ghc/ghc/-/issues/21575) is still not fixed as of GHC 9.4.1, I believe that issue only affects things like the `reflection` library, where the type of a class method does not uniquely determine the class dictionary. In `singletons`, however, the type of `sing` _does_ uniquely determine the `SingI` instance, so I think it would be perfectly safe to use `withDict` in `singletons`. (Please speak up if I've misunderstood the situation.)

goldfirere:
Your summary corresponds with my understanding. I think `singletons` is safe in this regard.

Issue #543, opened 2022-12-14
How does sCases work?


RobertFischer:
I'm trying to use `sCases`, but I'm constantly getting "GHC stage restriction".  How would I use this function to generate an implementation of a function?

If there's some documentation or a good example blog post that I'm missing, please point me in that direction.

RyanGlScott:
Indeed, the documentation for `sCases` (and [`cases`](https://hackage.haskell.org/package/singletons-th-3.1.1/docs/Data-Singletons-TH.html#v:cases), the function on which it is based) is rather lacking. A simple example of its use can be found [here](https://github.com/goldfirere/singletons/blob/70d622645b960d60411ab8f31d16ffb91e69379e/singletons-base/tests/compile-and-dump/Singletons/T124.hs). Suppose you have this:

```hs
module Singletons.T124 where

import Data.Singletons.TH
import Prelude.Singletons

$(singletons [d|
  foo :: Bool -> ()
  foo True = ()
  foo False = ()
  |])
```

And that you want to write a function of this type:

```hs
bar :: SBool b -> STuple0 (Foo b)
```

How would you do this? You might be tempted to write the following:

```hs
bar _ = STuple0
```

However, this won't typecheck, as `Foo b` won't reduce to `'()` unless GHC knows `b` is either `True` or `False`. In order to convince GHC of this, you must explicitly match on each of `SBool`'s data constructors:

```hs
bar :: SBool b -> STuple0 (Foo b)
bar b = case b of
  STrue  -> STuple0
  SFalse -> STuple0
```

This is doable, but somewhat tedious, and it only gets more tedious the more data constructors you have. For that reason, we offer `sCases` to generate this boilerplate code for you. The following is equivalent to the implementation of `bar` above:

```hs
bar :: SBool b -> STuple0 (Foo b)
bar b = $(sCases ''Bool [| b |] [| STuple0 |])
```

`cases` is a more general version of `sCases` that accepts any data type, not just singleton types. Note that `sCases ''Bool` is equivalent to `cases ''SBool`. (In a previous version of `singletons`, `cases` used to not be able to handle singleton types, which is why a separate `sCases` function was introduced. Nowadays, `sCases` mostly exists for historical reasons.)

RyanGlScott:
I've added the explanation above to the Haddocks in #544.

RobertFischer:
Thanks. That's super helpful.

Issue #527, opened 2022-08-18
Class for applying families


treeowl:
I may well be missing some utility or idiom, but today I found myself looking for something like

```haskell
class WDS (sym :: j ~> k) where
  wds :: forall t. Sing t -> Sing (sym @@ t)
```

I couldn't find anything, so I rolled my own. Is there a Right Way to approach this? If not, should there be?

RyanGlScott:
Does [`applySing`](https://hackage.haskell.org/package/singletons-3.0.1/docs/Data-Singletons.html#v:applySing) (or, if you like infix operators, [`(@@)`](https://hackage.haskell.org/package/singletons-3.0.1/docs/Data-Singletons.html#v:-64--64-)) give you what you seek?

treeowl:
A partially applied `applySing` looks like what I need, yes, but how to I get from `f` to `SLambda f`? I don't actually understand the role of `SLambda` in the infrastructure here.

RyanGlScott:
> I don't actually understand the role of SLambda in the infrastructure here.

Broadly speaking, `SLambda` is the singleton type for functions, just as `SBool` is the singleton type for `Bool`. Functions are treated a bit specially in `singletons`, which is why the actual definition of `SLambda` looks different from most singleton types, but it's really just a `Sing` function.

> A partially applied `applySing` looks like what I need, yes, but how to I get from `f` to `SLambda f`?

There are two ways of constructing an `SLambda` value:

1. With the `singFun*` family of combinators (e.g., [`singFun1`](https://hackage.haskell.org/package/singletons-3.0.1/docs/Data-Singletons.html#v:singFun1)). For example, if you have `sId :: Sing a -> Sing (Id a)`, then `singFun1 @IdSym0 sId :: SLambda IdSym0`, where `IdSym0 :: a ~> a` is the [defunctionalized](https://github.com/goldfirere/singletons/tree/489c1c8b685f36bdb9f1cbcbab52066009e044f8#promoting-functions) version of `Id`. 
2. Using the [`SingI`](https://hackage.haskell.org/package/singletons-3.0.1/docs/Data-Singletons.html#t:SingI) class. For example, `sing @IdSym0` is another way of constructing a value of type `SLambda IdSym0`. `singletons-th` automatically generates `SingI` instances for defunctionalization symbols such as `IdSym0`, which is why this technique works.

treeowl:
[...]

Issue #260, opened 2017-10-12
Could not deduce (SingI is)


LeanderK:
I am unable/unsure how to solve this problem:

```
Could not deduce (SingI is) arising from a use of ‘sing’
      from the context: Row row
        bound by the class declaration for ‘Row’
        at src/Extra/Row2.hs:104:7-9
```
`is` is of the kind `[Nat]`. 
Am i required to write the SingI instance by hand? Or can i generate it using singletons?
minimal non-compiling example (row is something like a optionally named record), i am currently experimenting with singletons, so don't expect great code 🙂

```haskell
$(promoteOnly [d|
    allInRange :: [Nat] -> [Type] -> Bool
    allInRange indexes types = all (\i -> i < leng) indexes
        where
            leng = length types

    mapToElement :: [Nat] -> [a] -> [a]
    mapToElement indexes types = map (\i -> types !! i) indexes

    maybeFilterNames :: [Nat] -> Maybe [Symbol] -> Maybe [Symbol]
    maybeFilterNames indexes (Just names) = Just $ map (\i -> names !! i) indexes
    maybeFilterNames indexes Nothing = Nothing

    getIndexOfNames :: [Symbol] -> Maybe [Symbol] -> Maybe [Nat]
    getIndexOfNames sys Nothing = Nothing
    getIndexOfNames sys (Just names) = seqMaybes $ map (\sy -> elemIndex sy names) sys

    seqMaybes :: [Maybe a] -> Maybe [a]
    seqMaybes [] = Just []
    seqMaybes ((Just x) : xs) = case seqMaybes xs of
        (Just rest) -> Just (x : rest)
        (Nothing)   -> Nothing
    seqMaybes ((Nothing) : ignored) = Nothing
 |])

type GetSliceNamedConstraints snames (is :: [Nat]) temp types names rts rns =
    (True ~ IsJust names, temp ~ GetIndexOfNames snames names, True ~ IsJust temp,
    is ~ FromJust temp, True ~ AllInRange is types, rts ~ MapToElement is types, 
    rns ~ MaybeFilterNames is names)

-- a row is ExampleRow (types :: [Type]) (names :: Maybe [Symbol])  

class Row row where
    getSlice :: 
        forall (is :: [Nat]) (rts :: [Type]) (rns :: Maybe [Symbol]) -- input / output types + names
            (types :: [Type]) (names :: Maybe [Symbol]). -- input row
[...]

PR #544, opened 2022-12-17
Improve documentation of `SingI`, `cases`, and `sCases`

RyanGlScott:
This PR is a collection of documentation improvements surrounding the `SingI` class (re #260 and #527), as well as the `cases`/`sCases` functions (re #543).

### DIFF
--- a/singletons-th/src/Data/Singletons/TH.hs
+++ b/singletons-th/src/Data/Singletons/TH.hs
@@ -71,8 +71,54 @@ import Language.Haskell.TH.Desugar
 -- is identical. This may be useful if the type-checker requires knowledge of which
 -- constructor is used to satisfy equality or type-class constraints, but where
 -- each constructor is treated the same.
+--
+-- Here is a simple example to illustrate where 'cases' can be useful. Suppose
+-- you use @singletons-th@ to single this code:
+--
+-- @
+-- $('singletons' [d|
+--   foo :: Bool -> ()
+--   foo True = ()
+--   foo False = ()
+--   |])
+-- @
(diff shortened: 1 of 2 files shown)


singletons-th/src/Data/Singletons/TH.hs
```diff
@@ -71,8 +71,54 @@ import Language.Haskell.TH.Desugar
...
+-- boilerplate code for you. The following is equivalent to the implementation
+-- of @bar@ above:
+--
+-- @
+-- bar :: SBool b -> STuple0 (Foo b)
+-- bar b = $(cases ''SBool [| b |] [| STuple0 |])
+-- @
 cases :: DsMonad q
-      => Name        -- ^ The head of the type of the scrutinee. (Like @''Maybe@ or @''Bool@.)
+      => Name        -- ^ The head of the type of the scrutinee. (Like @''Maybe@, @''Bool@, @''SBool@, etc.)
```
treeowl, line 121: Would the head of the scrutinee ever *not* be a GADTy type? Should the first two examples be removed?

treeowl:
Definitely better.

singletons-th/src/Data/Singletons/TH.hs
RyanGlScott: Yes, a fair point. Although nothing in the implementation of `cases` precludes the possibility of matching on a non-GADT type, it's difficult to imagine that one would ever want to do so. I've removed the mentions of `Maybe` and `Bool`.

goldfirere (approved):

PR #548, opened 2022-12-26
`singletons-th`: Remove unused `unSingFun*` machinery

### LINKED ISSUE #545: Does `singletons-th` ever actually generate code with `unSingFun*`?
While studying the code in `singletons-th` recently, I came across a peculiar comment in the Haddocks for `singClause`:

https://github.com/goldfirere/singletons/blob/570ca213a38fdeb4d77946a5861494bd96983571/singletons-th/src/Data/Singletons/TH/Single.hs#L749-L752

In particular, it mentions the possibility of eta expansion using `unSingFun`. This is strange, I couldn't find a single use of `unSingFun` in the entirety of the TH-generated code in `singletons-base`'s test suite. I also couldn't think of a way to come up with a custom example that would trigger the use of `unSingFun` either.

Looking at the code for `singClause`, I think I see why this is the case:

https://github.com/goldfirere/singletons/blob/570ca213a38fdeb4d77946a5861494bd96983571/singletons-th/src/Data/Singletons/TH/Single.hs#L771-L773

This computes `num_arrows` (i.e., the number of function arrows in the type of the clause's function) minus `length pats` (i.e., the number of patterns in the clause). You might expect that the former could be greater than the latter, e.g.:

```hs
myId :: a -> a
myId = id
```

I have come to the conclusion that by the time we reach `singClause`, however, we have ensured that there are exactly as many patterns as there are function arrows. This is because of `etaContractOrExpand`, which is used during function promotion:

https://github.com/goldfirere/singletons/blob/570ca213a38fdeb4d77946a5861494bd96983571/singletons-th/src/Data/Singletons/TH/Promote.hs#L900-L912

As a result, by the time we reach `singClause` (which occurs after promotion), we will have eta expanded the `myId` example above to:

```hs
myId :: a -> a
myId x = id x
```

As a result, I believe that `num_arrows` should always equal `length pats` in `singClause`. If this really is the case, it would be worth deleting all of the `wrapUnSingFun`-related code from `singClause`, as doing so should allow a modest amount of code simplification.

RyanGlScott:
Previous versions of `singletons` generated uses of `unSingFun*` as a way to accomplish eta expansion. In the current version of the library, however, eta expansion is instead accomplished by giving function clauses more patterns, which removes the need for `unSingFun*` entirely. There were still some vestigial remains of the old `unSingFun*` machinery lying around, however, so this patch removes them.

While I was in the neighborhood, I took the opportunity to document some of the invariants that the current eta expansion machinery relies on.

Fixes #545.

### DIFF
--- a/singletons-th/src/Data/Singletons/TH/Promote.hs
+++ b/singletons-th/src/Data/Singletons/TH/Promote.hs
@@ -780,7 +780,7 @@ promoteLetDecRHS rhs_sort type_env fix_env mb_let_uniq name let_dec_rhs = do
       expClauses <- mapM (etaContractOrExpand ty_num_args numArgs) clauses
       promote_let_dec_rhs all_locals m_ldrki ty_num_args
                           (mapAndUnzipM (promoteClause prom_fun_lhs) expClauses)
-                          id AFunction
+                          id (AFunction ty_num_args)
 
     -- Promote a UValue or a UFunction.
     -- Notes about type variables:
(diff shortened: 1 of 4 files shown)


goldfirere (approved):
LGTM

PR #549, opened 2022-12-27
`singletons-th`: Remove unneeded kind annotation when singling `case` expressions

### LINKED ISSUE #547: Do singled `case` expressions need an explicit kind annotation?
While pursuing a fix for #542, I uncovered this interesting piece of code:

https://github.com/goldfirere/singletons/blob/c419ece78017f1b4bcc22cd2427b7df21413ff9d/singletons-th/src/Data/Singletons/TH/Single.hs#L959-L965

In particular, the part of interest is the ``ret_ty `maybeSigT` res_ki`` bit. Why do we need the explicit `res_ki` kind annotation? That is, why do we need `id @(Sing (t :: k)) e` instead of just `id @(Sing t) e`? This is what `Note [Annotate case return type]` has to say:

https://github.com/goldfirere/singletons/blob/c419ece78017f1b4bcc22cd2427b7df21413ff9d/singletons-th/src/Data/Singletons/TH/Single.hs#L824-L837

Curiously, this Note never actually discusses the kind annotation portion of the code. As an experiment, I decided to change  ``ret_ty `maybeSigT` res_ki`` in the code to just `ret_ty`, and everything in `singletons-base` and its test suite continued to compile.

As historical context, this kind annotation was added in a60b00074f5a7f07b7e741e7daeb7f1b539e4121 in pursuit of a fix for #136. It's possible that this kind annotation was needed at the time, especially since `case` expressions were singled as `e :: Sing (t :: k)` rather than `id @(Sing (t :: k)) e`. This is because former interacts with kind generalization differently than the latter—see [`Note [The id hack; or, how singletons-th learned to stop worrying and avoid kind generalization]`](https://github.com/goldfirere/singletons/blob/570ca213a38fdeb4d77946a5861494bd96983571/singletons-th/src/Data/Singletons/TH/Single.hs#L1074-L1152) for the details. Now that `singletons-th` uses the `id` hack, it's quite possible that the kind annotation is no longer required due to the lack of kind generalization.

RyanGlScott:
Explicit kind annotations aren't necessary in modern `singletons-th` due to the approach that it uses for singling `case` expressions. This allows us to delete a modest amount of code.

Fixes #547.

### DIFF
--- a/singletons-base/tests/compile-and-dump/GradingClient/Database.golden
+++ b/singletons-base/tests/compile-and-dump/GradingClient/Database.golden
@@ -1444,7 +1444,7 @@ GradingClient/Database.hs:(0,0)-(0,0): Splicing declarations
                 sName'
         in
           (GHC.Base.id
-             @(Sing (Case_0123456789876543210 name name' u attrs (Let0123456789876543210Scrutinee_0123456789876543210Sym4 name name' u attrs) :: U)))
+             @(Sing (Case_0123456789876543210 name name' u attrs (Let0123456789876543210Scrutinee_0123456789876543210Sym4 name name' u attrs))))
             (case sScrutinee_0123456789876543210 of
                STrue -> sU
                SFalse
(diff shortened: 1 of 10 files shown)


goldfirere (approved):
LGTM!

PR #554, opened 2023-02-18
Monomorphize kinds of `Error{,WithoutStackTrace}`, introduce `PolyError`

### LINKED ISSUE #89: Extended error reporting
The error function tries to mimic the original one by accepting only Symbols. This significantly constrains error reporting because on the type level there is no show function and symbol manipulation.

I suggest to make its argument kind-polymorphic so that we can pass details on the error site.

RyanGlScott:
Previously, we had generalized the argument kind to `Error` in commit 077aee54a07f78fbff044a2a4801bfb992a898cf to permit passing things to `Error` besides `Symbol`s. Back then, there was no way to `show` things at the type level, nor was there a way to manipulate `Symbol`s in any meaningful fashion, so this seemed like a reasonable choice.

Nowadays, however, the story is different. There is a type-level `Show_` function, and the API for manipulating `Symbol`s is nearly as expressive as the API for manipulating `String`s.  What's more, making `Error`'s argument kind more general introduces ambiguity-related issues when deriving `Enum` instances with `OverloadedStrings` enabled, as observed in https://github.com/goldfirere/singletons/issues/89.

In light of this, I have changed the API such that:

* The kind of the argument to `Error` (as well as the related `ErrorWithoutStackTrace` function) is now `Symbol`.  In this sense, this patch reverts 077aee54a07f78fbff044a2a4801bfb992a898cf.
* There is now a new `Data.Singletons.Base.PolyError` module that provides a `PolyError` function. `PolyError` provides a kind-polymorphic `Error` interface much like what the previous type of `Error` was, so any existing code that relied on the argument of `Error` being kind-polymorphic can be migrated over to use `PolyError`.

Resolves #89 (hopefully for good this time).

### DIFF
--- a/singletons-base/CHANGES.md
+++ b/singletons-base/CHANGES.md
@@ -1,8 +1,18 @@
 Changelog for the `singletons-base` project
 ===========================================
 
-3.1.2 [????.??.??]
-------------------
+3.2 [????.??.??]
+----------------
+* The kinds of the promoted `Error` and `ErrorWithoutStackTrace` functions have
+  been monomorphized to `Symbol`. A previous release generalized the kinds of
+  these arguments to allow passing arguments besides `Symbol`s, but this change
+  introduces ambiguity in derived code when `OverloadedString`s is enabled.
+  See [#89](https://github.com/goldfirere/singletons/issues/89) for the full
+  story.
+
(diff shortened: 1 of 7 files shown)


amesgen (approved):
I think this is looking great, thanks for such a quick and thorough resolution!

Issue #566, opened 2023-06-16
Adapt to arity inference changes in type-level declarations


RyanGlScott:
Once [GHC!10660](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/10660) lands, GHC will no longer perform arity inference in type-level declarations, pursuant to [GHC proposal #425](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0425-decl-invis-binders.rst). Some code in the `singletons` libraries will need to be updated as a result. Here is a (possibly non-exhaustive) list, inspired by [`head.hackage!303`](https://gitlab.haskell.org/ghc/head.hackage/-/merge_requests/303):

* [ ] Currently, `singletons` defines the `Sing` type family like so:
  
  ```hs
  type Sing :: k -> Type
  type family Sing :: k -> Type
  ```

  (The reason that we have to duplicate the kind signature in the body is rather tiresome: it's because [Haddock doesn't support standalone kind signatures](https://github.com/haskell/haddock/issues/1178), so we need to include it for Haddocking purposes.)
  
  Currently, `Sing` is inferred to have arity 1, with one invisible binder for `k`. This will no longer compile after [GHC!10660](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/10660), so we will instead have to write it like this:
  
  ```hs
  type family Sing @k
  ```
* [ ] `TyCon1` is currently defined as:

   https://github.com/goldfirere/singletons/blob/2c91ce4ad7e80518f08ea7cf962c48820e83b309/singletons/src/Data/Singletons.hs#L632

   We will instead have to define it like this:

   ```hs
   type TyCon1 @k1 @k2 = (TyCon :: (k1 -> k2) -> (k1 ~> k2))
   ```

   And similarly for `TyCon2`, `TyCon3`, etc.
* [ ] "Fully saturated" defunctionalization symbols like this one will no longer kind-check:

   ```hs
   type MemptySym0 :: forall a. a
   type family MemptySym0 :: a where
     MemptySym0 = Mempty
   ```

   To make this work, we will instead have to generate code like this:

   ```hs
   type MemptySym0 :: forall a. a
   type family MemptySym0 @a :: a where
     MemptySym0 = Mempty
   ```

   See https://gitlab.haskell.org/ghc/head.hackage/-/merge_requests/303#note_505930 for a sketch of how to implement this.
* [ ] In a similar vein, several promoted type families will no longer kind-check, such as this one:

   ```hs
[...]

Issue #569, opened 2023-07-02
GHC 9.10 checklist


RyanGlScott:
This issue serves as a reminder to do certain things before we ship a new major release that supports GHC 9.10:

* [x] Update `th-desugar`. See https://github.com/goldfirere/th-desugar/issues/191.
* [x] Manually upload the Haddocks after uploading `singletons-th`/`singletons-base` to Hackage.
* [x] https://github.com/goldfirere/singletons/pull/584
* [x] #566

RyanGlScott:
* [`singletons-3.0.3`](https://hackage.haskell.org/package/singletons-3.0.3)
* [`singletons-th-3.4`](https://hackage.haskell.org/package/singletons-th-3.4)
* [`singletons-base-3.4`](https://hackage.haskell.org/package/singletons-base-3.4)

PR #587, opened 2024-05-04
Require building with GHC 9.10

RyanGlScott:
This is a collection of patches needed to make `singletons`, `singletons-th`, and `singletons-base` build with GHC 9.10 (and in the case of `singletons-{th,base}`, _require_ them to build with GHC 9.10). The two most interesting commits (whose commit messages I have included below) are the ones involving changes to the arities of type families. GHC 9.10 no longer performs arity inference (see the commit messages below), so we need to explicitly use `TypeAbstractions` in more places to give promoted type families the correct arities.

This makes more progress towards https://github.com/goldfirere/singletons/issues/569.

## `singletons`: Support building with GHC 9.10

Because GHC 9.10 no longer performs arity inference in type-level declarations (see
https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0425-decl-invis-binders.rst), we now must use `TypeAbstractions` in certain `singletons` definitions to make GHC 9.10 accept them.

One part of a fix for https://github.com/goldfirere/singletons/issues/566.

## `singletons-th`: Adapt to GHC 9.10's lack of arity inference

GHC 9.10 no longer performs arity inference in type-level declarations (see https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0425-decl-invis-binders.rst), and as it turns out, many of the promoted type families that `singletons-th` generates would have the wrong arity. For instance, `singletons-th` would promote this definition:

```hs
f :: Either a Bool
f = Right True
```

To this type family:

```hs
type F :: Either a Bool
type family F where
  F = Right True
```

With GHC 9.10 or later, however, GHC would conclude that `F` has arity 0, which means that it should not bind any arguments (visible or invisible). The type family equation for `F`, however, only works if `F` has arity 1! This is because the type family equation needs to bind an invisible `@a` argument:

```hs
  F @A = Right @A @Bool True
```

To ensure that type families like `F` have the expected arity, `singletons-th` now uses `TypeAbstractions` in more places to ensure that type family headers bind an appropriate number of type variables, which makes the type families' arities explicit. For instance, `singletons-th` now generates the following code for `F`:

```hs
type F :: Either a Bool
type family F @A where -- Note the @A here, which gives it arity 1
  F = Right True
```

For more details on how this is implemented, see the new `Note [Generating type families with the correct arity]` in `Data.Singletons.TH.Promote`.

A consequence of this change is that the average piece of `singletons-th`–generated code is much more likely to require `TypeAbstractions` than it did before. This explains why we now enable `TypeAbstractions` in almost every module in `singletons-base`.

Fixes https://github.com/goldfirere/singletons/issues/566.


singletons-base/tests/compile-and-dump/Singletons/PatternMatching.golden
```diff
@@ -594,7 +595,24 @@ Singletons/PatternMatching.hs:(0,0)-(0,0): Splicing declarations
...
+                                          GHC.Types.LiftedRep
+                                          ((:)
+                                             @GHC.Types.RuntimeRep
+                                             GHC.Types.LiftedRep
+                                             ('[] @GHC.Types.RuntimeRep))))’
+    • In a standalone kind signature for ‘Foo2’: (#,#) a b -> a
+   |
+19 | $(singletons [d|
+   |  ^^^^^^^^^^^^^^^...
+
```
RyanGlScott, line 618: This error message is unexpected. Upon further inspection, it's GHC fault, as `unboxedTupleTypeName 2` is returning the wrong `Name`, causing `singletons-th` to handle `(#,#)` in an unexpected way. See https://github.com/goldfirere/th-desugar/issues/213, as well as the upstream issue [GHC#24750](https://gitlab.haskell.org/ghc/ghc/-/issues/24750). Happily, that upstream GHC issue has already been fixed in `master`, so once that is backported to GHC 9.10, then this test case should work once more.

singletons-th/src/Data/Singletons/TH/Promote.hs
```diff
@@ -957,6 +966,56 @@ information whatsoever. Until the aformentioned Haddock issue is resolved, we
...
+
+Note [Generating type families with the correct arity]
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+As of GHC 9.10, GHC no longer performs arity inference when kind-checking type
+family declarations with standalone kind signatures. This is an important
+consideration when promoting functions with top-level type signatures. For
+example, we would not want to take this definition:
+
+  f :: Maybe a
+  f = Nothing
```
vagarenko, line 978: Did you mean to write 
```
f :: Either a Bool
f = Right True
```
here, like in the PR description?

RyanGlScott: Oops! I wrote this up, changed my mind about what example to use, and then forget to modify this part. Thanks for noticing this—I've changed it to be consistent with the rest of the documentation.

singletons-base/tests/compile-and-dump/Singletons/PatternMatching.golden
RyanGlScott: Happily, GHC 9.10.1 includes a fix for GHC#24750, so this is no longer an issue.

PR #598, opened 2024-05-31
Use explicit namespace specifiers when promoting/singling fixity declarations

### LINKED ISSUE #582: Support promoting/singling namespace specifiers in fixity declarations
GHC 9.10.1 [adds support for namespaces in fixity declarations](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/12012), e.g.,

```hs
$(singletons
  [d| infixl 4 data `f`
      f :: a -> a -> a
    |])
```

It would be nice if `singletons-th` could take these into account during promotion and singling. In particular, we could promote `f` to:

```hs
-- Promoted type-level declaration
infixl 4 type `F`
type F :: a -> a -> a
-- Defunctionalization symbols
infixl 4 type `FSym0`
infixl 4 type `FSym1`
```

And we could single `f` to:

```hs
infixl 4 data `sF`
sF :: forall a (x :: a) (y :: a). Sing x -> Sing y -> Sing (F x y)
```

Some wrinkles to this plan:

* What happens if we promote or single a quoted type-level declaration, such as this one?

  ```hs
  $(singletons
    [d| infixl 4 type `G`
        type G :: a -> a -> a
      |])
  ```

  In that case, we would simply return the original fixity declaration unchanged (including the `type` specifier).
* In the event that a fixity declaration lacks an explicit namespace specifier, I propose that we promote/single it to something that _does_ have an explicit namespace specifier. To see why, consider this example:

  ```hs
  $(singletons
    [d| infixl 4 ###
        (###) :: a -> a -> a
      |])
  ```
  
  We must _not_ promote this to the following:
  
  ```hs
  infixl 4 ###
  type (###) :: a -> a -> a
  ```
  
  If we did, there would be two clashing fixity declarations for `(###)`: one for the name in the `data` namespace, and another for the name in the `type` namespace. We should instead promote `(###)` to the following to avoid the risk of name clashes:
  
  ```hs
  infixl 4 type ###
  type (###) :: a -> a -> a
  ```
  
  If we did this, then we can remove a hack from `singletons-th` that is described in [`Wrinkle 1: When not to promote/single fixity declarations`](https://github.com/goldfirere/singletons/blob/2a604585d637d52256c31e1d4eee0c4d7150b7ef/singletons-th/src/Data/Singletons/TH/Single/Fixity.hs#L103-L116) of `Note [singletons-th and fixity declarations]`.

RyanGlScott:
During promotion, we now give promoted fixity declarations the `type` namespace, and during singling, we now give singled fixity declarations the `type` namespace (if it is a singled class declaration) or the `data` namesapce (for all other singled names). Doing so makes our intentions clearer, and it also future-proofs the code against planned GHC changes discussed in [GHC proposal 65](https://github.com/ghc-proposals/ghc-proposals/blob/8443acc903437cef1a7fbb56de79b6dce77b1a09/proposals/0065-type-infix.rst#migration-plan), where specifier-less fixity declarations (e.g., `infixl 4 %%%`) will become an error if they refer to both a term-level and type-level `%%%` name.

We now also take namespace specifiers into account when promoting fixity declarations. This is because `Wrinkle 1: When not to promote/single fixity declarations` in `Note [singletons-th and fixity declarations]` describes a special case where we must avoid promoting fixity declarations for infix names, but this workaround only needs to be used if the fixity declaration lacks an explicit namespace specifier. If it _does_ have an explicit namespace specifier (e.g., `infixl 4 data %%%`), then we can promote it (e.g., to `infixl 4 type %%%`) without fear of name clashes.

Fixes #582.

### DIFF
--- a/singletons-base/tests/compile-and-dump/Singletons/T582.golden
+++ b/singletons-base/tests/compile-and-dump/Singletons/T582.golden
@@ -1,0 +1,151 @@
+Singletons/T582.hs:(0,0)-(0,0): Splicing declarations
+    singletons
+      [d| infixl 4 !!!
+          infixl 4 %%%
+          infixl 4 `Bar`
+          infixl 4 `foo`
+          
+          foo :: a -> a -> a
+          x `foo` _ = x
+          (%%%) :: a -> a -> a
+          x %%% _ = x
+          
+          type Bar :: a -> a -> a
+          type (!!!) :: a -> a -> a
+          
+          type x `Bar` y = x
+          type x !!! y = x |]
+  ======>
+    infixl 4 `foo`
+    foo :: a -> a -> a
+    foo x _ = x
+    infixl 4 `Bar`
+    type Bar :: a -> a -> a
+    type Bar x y = x
+    infixl 4 %%%
+    (%%%) :: a -> a -> a
+    (%%%) x _ = x
+    infixl 4 !!!
+    type (!!!) :: a -> a -> a
+    type (!!!) x y = x
+    type BarSym0 :: (~>) a ((~>) a a)
+    data BarSym0 :: (~>) a ((~>) a a)
+      where
+        BarSym0KindInference :: SameKind (Apply BarSym0 arg) (BarSym1 arg) =>
+                                BarSym0 a0123456789876543210
+    type instance Apply BarSym0 a0123456789876543210 = BarSym1 a0123456789876543210
+    instance SuppressUnusedWarnings BarSym0 where
+      suppressUnusedWarnings = snd ((,) BarSym0KindInference ())
+    infixl 4 `BarSym0`
+    type BarSym1 :: a -> (~>) a a
+    data BarSym1 (a0123456789876543210 :: a) :: (~>) a a
+      where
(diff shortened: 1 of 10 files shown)


singletons-base/tests/compile-and-dump/Singletons/T582.golden
```diff
@@ -0,0 +1,151 @@
...
+    type (%%%) :: a -> a -> a
+    type family (%%%) @a (a :: a) (a :: a) :: a where
+      (%%%) x _ = x
+    type Foo :: a -> a -> a
+    type family Foo @a (a :: a) (a :: a) :: a where
+      Foo x _ = x
+    infixl 4 %%%
+    infixl 4 `Foo`
+    infixl 4 %%%%
+    infixl 4 `sFoo`
```
RyanGlScott, line 128: I was surprised to see that these generated fixity declarations seemingly lacked namespace specifiers. Some investigation reveals that the declarations actually _do_ have namespace specifiers, but due to a [GHC pretty-printing bug](https://gitlab.haskell.org/ghc/ghc/-/issues/24911), the namespaces don't get printed in the `-ddump-splices` output. Thankfully, the generated code still behaves as expected, but we'll have to live with some counterintuitive golden output until that GHC bug gets fixed upstream.

Issue #601, opened 2024-06-01
Make `singletons` buildable after GHC#23515


RyanGlScott:
[GHC#23515](https://gitlab.haskell.org/ghc/ghc/-/issues/23515) tracks implementing [this portion of GHC proposal #425](https://github.com/ghc-proposals/ghc-proposals/blob/8443acc903437cef1a7fbb56de79b6dce77b1a09/proposals/0425-decl-invis-binders.rst#proposed-change-specification-instances) (Invisible binders in type declarations):

> In type family and data family instances, the instantiation is fully determined by the left hand side, without looking at the right hand side.

A reasonable change, but one that will nevertheless cause several parts of `singletons` and `singletons-base` to fail to compile. This issue aims to identify all of these sources of breakage and how to fix them.

## `Sing` instances without explicit left-hand sides

There are places in `singletons` and `singletons-base` that declare `Sing` instances without specifying what type the `Sing` instance is for on the left-hand side, e.g.,

```hs
type instance Sing = SNat
```

This will no longer work after [GHC#23515](https://gitlab.haskell.org/ghc/ghc/-/issues/23515) is implemented. Instead, we must write:

```hs
type instance Sing @Nat = SNat
```

The following places in the code will need to be updated:

* https://github.com/goldfirere/singletons/blob/7e5c92d29f100e9a56f64854abbe35af1119ecac/singletons/src/Data/Singletons.hs#L408
* https://github.com/goldfirere/singletons/blob/7e5c92d29f100e9a56f64854abbe35af1119ecac/singletons/src/Data/Singletons.hs#L733
* https://github.com/goldfirere/singletons/blob/7e5c92d29f100e9a56f64854abbe35af1119ecac/singletons/src/Data/Singletons/Sigma.hs#L102
* https://github.com/goldfirere/singletons/blob/7e5c92d29f100e9a56f64854abbe35af1119ecac/singletons/tests/ByHand.hs#L101
* https://github.com/goldfirere/singletons/blob/7e5c92d29f100e9a56f64854abbe35af1119ecac/singletons/tests/ByHand.hs#L155
* https://github.com/goldfirere/singletons/blob/7e5c92d29f100e9a56f64854abbe35af1119ecac/singletons/tests/ByHand.hs#L195
* https://github.com/goldfirere/singletons/blob/7e5c92d29f100e9a56f64854abbe35af1119ecac/singletons/tests/ByHand.hs#L245
[...]

PR #610, opened 2024-06-25
Record local variables' kinds during lambda lifting

RyanGlScott:
Previously, `singletons-th` made no effort to track the kinds of local variables when generating lambda-lifted code, instead generating local variable binders with no kind annotations. As a result, GHC would generalize the kinds of these lambda-lifted definitions to things that are way more polymorphic than intended. While this technically works in today's GHC, it won't in a future version of GHC that implements [GHC#23515](https://gitlab.haskell.org/ghc/ghc/-/issues/23515).

In general, generating kinds for every local variable would require `singletons-th` to implement something akin to full-blown type inference over the Template Haskell AST, which is not something I am eager to implement.

Fortunately, there is a relatively simple approach we can do to alleviate this problem that doesn't require full type inference. In situations where we know the kind of a local variable (e.g., when there is a top-level signature or there is a pattern signature), we record the variable's kind and use it when generating binders for any lambda-lifted definitions that close over the variable. For the full story on how this works, see `Note [Local variables and kind information]` `D.S.TH.Promote.Syntax.LocalVar`.

This is not a perfect solution, as there will still be examples of the original problem that won't be covered by this simple approach (see the Note). This approach is still much better than what `singletons-th` was doing before, and I think it's worth using this simple approach even if it doesn't fix 100% of all cases.

This patch mostly resolves the "Overly polymorphic lambda-lifting, part 2" section of #601.


RyanGlScott:
I've marked this as a draft because I still need to test these changes with a version of GHC that is built with the changes from [GHC!12776](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/12776). My hope is that this patch will be enough to fix the breakages that I noticed in https://github.com/goldfirere/singletons/issues/601#overly-polymorphic-lambda-lifting-part-2, but I need to check.

RyanGlScott:
I've verified that this patch allows all of the examples in https://github.com/goldfirere/singletons/issues/601#overly-polymorphic-lambda-lifting-part-2 to compile with a version of GHC that is built with the changes from [GHC!12776](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/12776). As such, I think this is ready to land.

PR #620, opened 2024-12-03
`singletons-{th,base}`: Require building with GHC 9.12

RyanGlScott:
The most notable change is that because GHC 9.12 includes a fix for https://gitlab.haskell.org/ghc/ghc/-/issues/24911, the test suite output now includes explicit namespaces for its fixity declarations. Aside from that, the changes here are mostly bumping upper version bounds.

Checks off one box in https://github.com/goldfirere/singletons/issues/619.

PR #621, opened 2024-12-05
Support ghc 9.12

erikd:


RyanGlScott:
I'd prefer to do this as part of the larger changes in #620, so I'll close this PR in favor of that one.

PR #629, opened 2026-01-30
Create deterministic names for all identifiers and variables

mpickering:
@RyanGlScott If you had the time I would appreciate your feedback on the approach in this patch before I invest time finishing things off.

```
There are currently two sources of non-determinism when using
singletons-th.

1. Let bindings are promoted to a top-level type family named
   Let<UNIQUE>.

2. The `noExactTyVars` function shifts the unique into the OccName of a
   name in order to make distinct occnames.

Now there are two mechanisms to provide deterministic names in these two
cases.

1. The PrM monad carries an incrementing counter in order to uniquely
   name each top-level Let. Any name created by `newUniqueName` now uses
   this global counter.

2. For the names which require local uniqueness, `noExactTyVars` carries
   its own counter and renaming in order to substitute fresh names into
   a local scope.

Fixes https://github.com/goldfirere/singletons/issues/628
```

### DIFF
--- a/singletons-th/src/Data/Singletons/TH/Util.hs
+++ b/singletons-th/src/Data/Singletons/TH/Util.hs
@@ -7,7 +7,10 @@ This file contains helper functions internal to the singletons-th package.
 Users of the package should not need to consult this file.
 -}
 
-module Data.Singletons.TH.Util ( module Data.Singletons.TH.Util, UniqueCounter(..) ) where
+module Data.Singletons.TH.Util
+  ( module Data.Singletons.TH.Util
+  , UniqueCounter(..)
+  ) where
 
 import Prelude hiding ( exp, foldl, concat, mapM, any, pred )
 import Language.Haskell.TH ( pprint )
@@ -384,23 +387,28 @@ filterInvisTvbArgs (DFAForalls tele args) =
     DForallVis   _     -> res
     DForallInvis tvbs' -> tvbs' ++ res
 
--- | Change all unique Names with a NameU or NameL namespace to non-unique Names
--- by performing a syb-based traversal. See Note [Pitfalls of NameU/NameL] for
--- why this is useful.
+-- | Change all unique 'Name's with a 'NameU' or 'NameL' namespace to non-unique
+-- 'Name's by performing a syb-based traversal. See @Note [Pitfalls of
+-- NameU/NameL]@ for why this is useful.
 --
--- Each variable encountered during the traversal will be given a fresh non-unique Name
--- which is the variable OccName followed by an incrementing counter.
--- For example: x0 y1 z2 x3 etc
+-- Each variable encountered during the traversal will be given a fresh
+-- non-unique 'Name' which is the variable OccName followed by an incrementing
+-- counter. For example: @x0@, @y1@, @z2@, @x3@, etc.
 --
--- Each call to `noExactTyVars` starts producing names from 0, so
+-- Each call to 'noExactTyVars' starts producing names from @0@. As a result,
+-- if you call 'noExactTyVars' in two different locations, then there is no
+-- guarantee that the 'Name's used in one location will be distinct from the
+-- other location. It is the responsibility of the caller to ensure that this
+-- does not result in name clashes.
 noExactTyVars :: Data a => a -> a
 noExactTyVars = runFreshen . freshenTyVarsM
 
--- | This function performs an SYB traversal and replaces `Name`s with a fresh name,
--- and substitutes the fresh name for all other occurences of the name in the expression.
+-- | This function performs an SYB traversal and replaces `Name`s with a fresh
+-- name, and substitutes the fresh name for all other occurences of the name in
+-- the expression.
 --
--- The scope of 'freshenTyVarsM' is indicated by executing the stateful computation using
--- `runFreshen`.
+-- The scope of 'freshenTyVarsM' is indicated by executing the stateful
+-- computation using 'runFreshen'.
 freshenTyVarsM :: Data a => a -> State FreshenState a
 freshenTyVarsM = everywhereM freshenTyVarsStep
   where
@@ -435,17 +443,22 @@ freshenTyVarsM = everywhereM freshenTyVarsStep
       n' <- freshenName n
       pure LocalVar { lvName = n', lvKind = mbKind }
 
+-- | The state maintained by 'freshenTyVarsM'.
 data FreshenState = FreshenState
-  { fsNextId   :: !Int
+  { fsNextId    :: !Int
   , fsRenamings :: Map Name Name
   }
 
+-- | The initial state for 'freshenTyVarsM': identifiers will start from @0@,
+-- and no renamings will have yet occurred.
 initialFreshenState :: FreshenState
 initialFreshenState = FreshenState 0 Map.empty
 
+-- | Run a stateful computation that involves 'FreshenState'.
 runFreshen :: State FreshenState a -> a
 runFreshen = (`evalState` initialFreshenState)
 
+-- | Create a fresh name and update the 'FreshenState' accordingly.
 freshenName :: Name -> State FreshenState Name
 freshenName n@(Name (OccName occ) ns) =
   case ns of
--- a/singletons-th/src/Data/Singletons/TH/Util.hs
+++ b/singletons-th/src/Data/Singletons/TH/Util.hs
@@ -7,7 +7,7 @@ This file contains helper functions internal to the singletons-th package.
 Users of the package should not need to consult this file.
 -}
 
-module Data.Singletons.TH.Util where
+module Data.Singletons.TH.Util ( module Data.Singletons.TH.Util, UniqueCounter(..) ) where
 
 import Prelude hiding ( exp, foldl, concat, mapM, any, pred )
 import Language.Haskell.TH ( pprint )
@@ -19,6 +19,7 @@ import Control.Monad ( liftM, unless, when )
 import Control.Monad.Except ( ExceptT, runExceptT, MonadError(..) )
 import Control.Monad.IO.Class ( MonadIO )
 import Control.Monad.Reader ( MonadReader(..), Reader, ReaderT(..) )
+import Control.Monad.State.Strict ( State, evalState, get, modify' )
 import Control.Monad.Trans ( MonadTrans )
 import Control.Monad.Writer ( MonadWriter(..), WriterT(..), execWriterT )
 import qualified Data.Map as Map
@@ -38,14 +39,6 @@ qReportWarning = qReport False
 -- like reportError, but generalized to any Quasi
 qReportError :: Quasi q => String -> q ()
 qReportError = qReport True
-
--- | Generate a new Unique
-qNewUnique :: DsMonad q => q Uniq
-qNewUnique = do
-  Name _ flav <- qNewName "x"
-  case flav of
-    NameU n -> return n
-    _       -> error "Internal error: `qNewName` didn't return a NameU"
 
 checkForRep :: Quasi q => [Name] -> q ()
 checkForRep names =
@@ -168,31 +161,6 @@ tailNameStr str =
   case str of
     (_:cs) -> cs
     [] -> error "tailNameStr: Expected non-empty string"
-
--- convert a number into both alphanumeric and symoblic forms
-uniquePrefixes :: String   -- alphanumeric prefix
-               -> String   -- symbolic prefix
-               -> Uniq
-               -> (String, String)  -- (alphanum, symbolic)
-uniquePrefixes alpha symb n = (alpha ++ n_str, symb ++ convert n_str)
-  where
-    n_str = show n
-
-    convert [] = []
-    convert (d : ds) =
-      let d' = case d of
-                 '0' -> '!'
-                 '1' -> '#'
-                 '2' -> '$'
-                 '3' -> '%'
-                 '4' -> '&'
-                 '5' -> '*'
-                 '6' -> '+'
-                 '7' -> '.'
-                 '8' -> '/'
-                 '9' -> '>'
-                 _   -> error "non-digit in show #"
-      in d' : convert ds
 
 -- extract the kind from a TyVarBndr
 extractTvbKind :: DTyVarBndr flag -> Maybe DKind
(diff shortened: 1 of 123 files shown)


singletons-th/src/Data/Singletons/TH/Deriving/Eq.hs
```diff
@@ -18,7 +18,7 @@ import Data.Singletons.TH.Deriving.Util
 import Data.Singletons.TH.Names
 import Data.Singletons.TH.Syntax
 import Data.Singletons.TH.Util
-import Language.Haskell.TH.Desugar
+import Language.Haskell.TH.Desugar hiding (newUniqueName)
```
RyanGlScott: Given its issues with non-deterministic output, I wonder if we should deprecate (and eventually remove) `newUniqueName` in `th-desugar`.

singletons-th/src/Data/Singletons/TH/Util.hs
```diff
@@ -169,30 +162,35 @@ tailNameStr str =
...
+newtype UniqueId = UniqueId { unUniqueId :: Int }
+  deriving (Eq, Ord, Show)
+
+-- | Produce a deterministic identifier for naming generated bindings.
+nextUniqueId :: Quasi q => q UniqueId
+nextUniqueId = do
+  mCounter <- qGetQ
+  let DeterministicCounter current =
+        maybe (DeterministicCounter 0) id mCounter
+      next = current + 1
```
RyanGlScott: I wonder if this should fail if this exceeds the maximum value of an `Int`? I suspect that arbitrarily bad things might happen if you overflow and start re-using unique IDs.

singletons-th/src/Data/Singletons/TH/Util.hs
```diff
@@ -169,30 +162,35 @@ tailNameStr str =
...
--- convert a number into both alphanumeric and symoblic forms
-uniquePrefixes :: String   -- alphanumeric prefix
-               -> String   -- symbolic prefix
-               -> Uniq
-               -> (String, String)  -- (alphanum, symbolic)
-uniquePrefixes alpha symb n = (alpha ++ n_str, symb ++ convert n_str)
+newtype DeterministicCounter = DeterministicCounter Int
+
+newtype UniqueId = UniqueId { unUniqueId :: Int }
+  deriving (Eq, Ord, Show)
```
RyanGlScott: A couple of questions related to the data representation:

* Should we use `Word` instead of `Int` here? I don't think that we'd ever want to allow for the possibility of having a negative unique ID.
* Similarly, should we use `Word64` instead of just `Word`? GHC itself made the choice to use 64-bit unique IDs on 32-bit platforms (see [GHC#22010](https://gitlab.haskell.org/ghc/ghc/-/issues/22010)), and I wonder if we should do the same here.

singletons-th/src/Data/Singletons/TH/Util.hs
```diff
@@ -169,30 +162,35 @@ tailNameStr str =
...
-                 '6' -> '+'
-                 '7' -> '.'
-                 '8' -> '/'
-                 '9' -> '>'
-                 _   -> error "non-digit in show #"
-      in d' : convert ds
+    sanitize c
+      | c == '.'  = '_'
+      | c == '\'' = '_'
+      | otherwise = c
```
RyanGlScott: What about names with symbolic (infix) characters? Surely these would need to be sanitized as well? (See the only `convert` code from `uniquePrefixes`.)

singletons-th/src/Data/Singletons/TH/Util.hs
```diff
@@ -419,41 +417,76 @@ filterInvisTvbArgs (DFAForalls tele args) =
 -- Change all unique Names with a NameU or NameL namespace to non-unique Names
 -- by performing a syb-based traversal. See Note [Pitfalls of NameU/NameL] for
 -- why this is useful.
+--
+-- Each variable encountered during the traversal will be given a fresh non-unique Name
+-- which is the variable OccName followed by an incrementing counter.
+-- For example: x0 y1 z2 x3 etc
 noExactTyVars :: Data a => a -> a
```
RyanGlScott: If I understand correctly, then if you call `noExactTyVars` in two different locations (e.g., `(noExactTyVars x, noExactTyVars y)`), then there is no guarantee that the `Name`s used in one location will be distinct from the names used the other location. Is that correct? If so, it would be worth documenting this potential pitfall.

As an alternative, we could remove `noExactTyVars` and instead use `freshenTyVarsM` everywhere. I don't think this should be too hard to do—there's only one `noExactTyVars` call site remaining, and I suspect that we could replace it with `freshenTyVarsM` by making the function which calls it monadic.

singletons-th/src/Data/Singletons/TH/Util.hs
RyanGlScott: Should this incrementing step be evaluated strictly? I worry about the chance for thunk buildup here.

mpickering:
Thanks for your review @RyanGlScott I will return to this tomorrow. I've finished chasing down the other determinism issues for now!

singletons-th/src/Data/Singletons/TH/Deriving/Eq.hs
mpickering: That would be a good idea, we would like the ecosystem to move towards deterministic builds.

singletons-th/src/Data/Singletons/TH/Util.hs
mpickering: Yes, bad things could happen but you probably would run into other issues trying to compile a program with so many top-level clauses! I have changed it to `Word64` to be on the safe side.

singletons-th/src/Data/Singletons/TH/Util.hs
mpickering: I changed it to `Word64`.

singletons-th/src/Data/Singletons/TH/Util.hs
mpickering: I think if you look at the callsites for `newUniqueId`, it's always called with a string which is is alphanumeric (prefix). Therefore the sanitising is just important for replacing characters in module names.

singletons-th/src/Data/Singletons/TH/Util.hs
mpickering: Yes. that's right, fresh names are created within each scope and substituted into the remaining parts of the expression.

singletons-th/src/Data/Singletons/TH/Util.hs
mpickering: Done.

singletons-th/src/Data/Singletons/TH/Util.hs
mpickering: The place where `noExactTyVars` is called takes care to more precisely substitute into the right places
rather than the code in `Data.Singletons.TH.Promote.Defun` which also uses the traversal to perform the substitution. I feel like all this code is a bit hairy and I don't fully understand what it's doing so I would prefer to not modify it too intensively unless there are some additional tests which demonstrate what's missing. Is that ok with you Ryan?

mpickering:
I've made some of the changes you suggest @RyanGlScott. I'm not so familar with this code so I'm quite wary to embark on any major refactorings, perhaps this would be a good interim step and provide a baseline for improving the name freshening story at a later point.

RyanGlScott:
Thanks! (Did you mean to push your changes? I see that various discussion threads have been closed, but the code in the PR itself hasn't yet changed.)

On the topic of removing `th-desugar`'s `newUniqueName` function, I have realized that things aren't quite as simple as I originally envisioned. In particular, the `th-desugar` library itself uses `newUniqueName` when desugaring code. What's more, `singletons-th` is built on top of this desugaring functionality, which means that even after all of the changes in this PR, I suspect that the output of `singletons-th` is _still_ not deterministic.

As a result, I am wondering if a better approach would be to port the unique identifier machinery from this PR over to `th-desugar`, and then we could update `singletons-th` to build on top of that machinery. I've opened https://github.com/goldfirere/th-desugar/issues/239 to track this.

What do you think about this idea? It would unfortunately take some more work to implement, but if our goal is having a high degree of confidence that the output is deterministic, then I think we'll need to do something like this. (If you don't want to do this, that's fine too—I can understand if you feel like this has been more work than you originally anticipated.)

singletons-th/src/Data/Singletons/TH/Util.hs
RyanGlScott: That's fine! I am happy to take a closer look at this myself in a follow-up PR.

singletons-th/src/Data/Singletons/TH/Util.hs
RyanGlScott: Ah, fair enough. And now that I look more closely, the Haddocks for [`th-desugar`'s `newUniqueName`](https://hackage-content.haskell.org/package/th-desugar-1.19/docs/Language-Haskell-TH-Desugar.html#v:newUniqueName) (on which this function is based) do mention that the string has to be alphanumeric. Perhaps it would suffice to just document that here as well.

mpickering:
@RyanGlScott I pushed now, I accidentally pushed to the wrong remote before.

I'm able to fix the level at whichever level is best for the ecosystem. For the project which I tested this on, these changes were sufficient. I am currently trying to modify the test harness to see if it can test for these kinds of failures so that we don't regress in the future if something changes.

mpickering:
I've modified the testsuite now to check for deterministic interface files for all the tests. This has highlighted there are still some issues, perhaps that is your observation about the `th-desugar` issue being the root cause of the non-determinism.

RyanGlScott:
Thanks for being extra thorough here! The determinism-related tests failing matches my expectations, as I suspect that every use of `newUniqueName` in `th-desugar` is a potential source of non-determinism. In particular, I think that the uses in [`Language.Haskell.TH.Desugar`](https://github.com/goldfirere/th-desugar/blob/fb1a3b142d15c8c2bc38372da7c0698ec9a6906d/Language/Haskell/TH/Desugar/Core.hs) are the most dangerous, as these are involved any time you desugar the `template-haskell` AST to the `th-desugar` AST that `singletons-th` is built on top of.

As a particularly simple example, the use of `newUniqueName` [here](https://github.com/goldfirere/th-desugar/blob/fb1a3b142d15c8c2bc38372da7c0698ec9a6906d/Language/Haskell/TH/Desugar/Core.hs#L1182) is particularly bad, as that line of code gets invoked every time you desugar any function clause with a `where` clause. As an example, consider:

```hs
{-# LANGUAGE GHC2024 #-}
{-# LANGUAGE TypeAbstractions #-}
{-# LANGUAGE TypeFamilies #-}
module Bug where

import Data.Singletons.TH

$(singletons [d|
  f :: a -> a
  f x = y
    where
      y = x
  |])
```

You'll see different `-ddump-splices` output depending on how GHC's unique numbers are incremented:

```
$ ghc-9.14 Bug.hs -ddump-splices -fforce-recomp -dunique-increment=-1
Loaded package environment from /Users/rscott/Documents/Hacking/Haskell/singletons/.ghc.environment.aarch64-darwin-9.14.1
[1 of 1] Compiling Bug              ( Bug.hs, Bug.o )
Bug.hs:(10,2)-(15,5): Splicing declarations
    singletons
      [d| f_a5k1wlNFEup :: a_a5k1wlNFEuo -> a_a5k1wlNFEuo
          f_a5k1wlNFEup x_a5k1wlNFEun
            = y_a5k1wlNFEum
            where
                y_a5k1wlNFEum = x_a5k1wlNFEun |]
  ======>
    f_a5k1wlNFEiL :: a_a5k1wlNFEiM -> a_a5k1wlNFEiM
    f_a5k1wlNFEiL x_a5k1wlNFEiK
      = y_a5k1wlNFEiJ
      where
          y_a5k1wlNFEiJ = x_a5k1wlNFEiK
    type family Let7061644215716926668YSym0 (x7061644215716926669 :: a7061644215716926676) where
      Let7061644215716926668YSym0 x7061644215716926669 = Let7061644215716926668Y x7061644215716926669
    type family Let7061644215716926668Y (x7061644215716926669 :: a7061644215716926676) where
      Let7061644215716926668Y x_a5k1wlNFEiF = x_a5k1wlNFEiF
    type FSym0 :: (~>) a_a5k1wlNFEiM a_a5k1wlNFEiM
    data FSym0 :: (~>) a_a5k1wlNFEiM a_a5k1wlNFEiM
      where
        FSym0KindInference :: SameKind (Apply FSym0 arg_a5k1wlNFEiH) (FSym1 arg_a5k1wlNFEiH) =>
                              FSym0 a7061644215716926670
    type instance Apply @a_a5k1wlNFEiM @a_a5k1wlNFEiM FSym0 a7061644215716926670 = F a7061644215716926670
    instance SuppressUnusedWarnings FSym0 where
      suppressUnusedWarnings = snd ((,) FSym0KindInference ())
    type FSym1 :: a_a5k1wlNFEiM -> a_a5k1wlNFEiM
    type family FSym1 @a_a5k1wlNFEiM (a7061644215716926670 :: a_a5k1wlNFEiM) :: a_a5k1wlNFEiM where
      FSym1 a7061644215716926670 = F a7061644215716926670
    type F :: a_a5k1wlNFEiM -> a_a5k1wlNFEiM
    type family F @a_a5k1wlNFEiM (a_a5k1wlNFEiI :: a_a5k1wlNFEiM) :: a_a5k1wlNFEiM where
      F x_a5k1wlNFEiF = Let7061644215716926668YSym0 x_a5k1wlNFEiF
    sF ::
      (forall (t_a5k1wlNFEiC :: a_a5k1wlNFEiM).
       Sing t_a5k1wlNFEiC
       -> Sing (F t_a5k1wlNFEiC :: a_a5k1wlNFEiM) :: Type)
    sF (sX :: Sing x_a5k1wlNFEiF)
      = let
          sY :: Sing @_ (Let7061644215716926668Y x_a5k1wlNFEiF)
          sY = sX
        in sY
    instance SingI (FSym0 :: (~>) a_a5k1wlNFEiM a_a5k1wlNFEiM) where
      sing = singFun1 @FSym0 sF
```
```
$ ghc-9.14 Bug.hs -ddump-splices -fforce-recomp -dunique-increment=1
Loaded package environment from /Users/rscott/Documents/Hacking/Haskell/singletons/.ghc.environment.aarch64-darwin-9.14.1
[1 of 1] Compiling Bug              ( Bug.hs, Bug.o )
Bug.hs:(10,2)-(15,5): Splicing declarations
    singletons
      [d| f_a2GD :: a_a2GE -> a_a2GE
          f_a2GD x_a2GF
            = y_a2GG
            where
                y_a2GG = x_a2GF |]
  ======>
    f_a2Sh :: a_a2Sg -> a_a2Sg
    f_a2Sh x_a2Si
      = y_a2Sj
      where
          y_a2Sj = x_a2Si
    type family Let6989586621679020852YSym0 (x6989586621679020851 :: a6989586621679020844) where
      Let6989586621679020852YSym0 x6989586621679020851 = Let6989586621679020852Y x6989586621679020851
    type family Let6989586621679020852Y (x6989586621679020851 :: a6989586621679020844) where
      Let6989586621679020852Y x_a2Sn = x_a2Sn
    type FSym0 :: (~>) a_a2Sg a_a2Sg
    data FSym0 :: (~>) a_a2Sg a_a2Sg
      where
        FSym0KindInference :: SameKind (Apply FSym0 arg_a2Sl) (FSym1 arg_a2Sl) =>
                              FSym0 a6989586621679020850
    type instance Apply @a_a2Sg @a_a2Sg FSym0 a6989586621679020850 = F a6989586621679020850
    instance SuppressUnusedWarnings FSym0 where
      suppressUnusedWarnings = snd ((,) FSym0KindInference ())
    type FSym1 :: a_a2Sg -> a_a2Sg
    type family FSym1 @a_a2Sg (a6989586621679020850 :: a_a2Sg) :: a_a2Sg where
      FSym1 a6989586621679020850 = F a6989586621679020850
    type F :: a_a2Sg -> a_a2Sg
    type family F @a_a2Sg (a_a2Sk :: a_a2Sg) :: a_a2Sg where
      F x_a2Sn = Let6989586621679020852YSym0 x_a2Sn
    sF ::
      (forall (t_a2Sq :: a_a2Sg).
       Sing t_a2Sq -> Sing (F t_a2Sq :: a_a2Sg) :: Type)
    sF (sX :: Sing x_a2Sn)
      = let
          sY :: Sing @_ (Let6989586621679020852Y x_a2Sn)
          sY = sX
        in sY
    instance SingI (FSym0 :: (~>) a_a2Sg a_a2Sg) where
      sing = singFun1 @FSym0 sF
```

mpickering:
Thanks Ryan, I added this test and it passes the determinism check on the branch.

Here is a minimised version of one of the failures:

```
$(singletons [d|
  foo7 (x :: a) (_ :: b) = x
  |])
```

which results in

```
      -    forall a (x :: a) b (wild_6989586621679030148 :: b).
      -      Foo7 x wild_6989586621679030148 = x
      +    forall a (x :: a) b (wild_6989586621680989414 :: b).
      +      Foo7 x wild_6989586621680989414 = x
```

mpickering:
Updating `th-desugar` seems to fix all the issues with determinism :pray: 

I will open a PR over there.

mpickering:
@RyanGlScott I pushed another update which should be passing CI now (with the updated th-desugar commit).

singletons-th/src/Data/Singletons/TH/Util.hs
```diff
@@ -169,30 +163,20 @@ tailNameStr str =
...
+
+-- | Produce a deterministic identifier for naming generated bindings.
+nextUniqueId :: Quasi q => q UniqueId
+nextUniqueId = do
+  mCounter <- qGetQ
+  let DeterministicCounter current =
+        maybe (DeterministicCounter 0) id mCounter
+      !next = current + 1
+  qPutQ (DeterministicCounter next)
+  pure (UniqueId current)
```
RyanGlScott: This code is remarkably similar to the internals of `newUniqueName`'s implementation in https://github.com/goldfirere/th-desugar/pull/240, to the point that I wonder if we should move `nextUniqueId` to `th-desugar` and define `newUniqueName` in terms of `nextUniqueId`. I like this idea, as it would allow us to keep all of the unique ID implementation within `th-desugar`, and then all `singletons-th` would need to do is import it.

RyanGlScott (commented):
Fantastic work! At this point, I think my only question concerns which parts of the code should live in `th-desugar` versus `singletons-th`.

mpickering:
@RyanGlScott I have pushed the changes to `th-desugar` and updated/tested this PR against those changes. Let me know if there is anything else you would like me to do.

singletons-th/src/Data/Singletons/TH/Util.hs
RyanGlScott: This was done in https://github.com/goldfirere/th-desugar/pull/240.

RyanGlScott (approved):
Thanks again, @mpickering! Let me know if you need new releases of `singletons-{th,base}` soon with these changes.

